PR #13, opened 2026-03-24
feat: 3.2 order intake checks — max order rate and duplicate window hardening

blahmonkey:

## What does this PR change?

Implements RFC 3.2 intake-check hardening across configuration, hook wiring, and deterministic validation coverage. Adds `InstanceConfig.max_order_rate` with strict runtime guards (reject `bool`, non-`int`, non-positive values). Updates default intake hook assembly to source `MAX_ORDER_RATE` from config while preserving explicit override precedence. Expands intake-stage tests to lock idempotency semantics on (`strategy_id`, `command_id`), verify temporal window behavior (rate-limit recovery and duplicate replay expiry), enforce non-ENTER duplicate bypass behavior, and stabilize deny reason message assertions. Updates changelog and bumps package version to v0.12.0.

## Checklist

- [x] I ran `python -m pytest` locally without errors (470 passing)
- [x] I updated CHANGELOG.md

### DIFF
--- a/nexus/core/validator/intake_stage.py
+++ b/nexus/core/validator/intake_stage.py
@@ -43,7 +43,8 @@ def build_default_intake_hooks(
     Args:
         config: Instance config carrying duplicate window settings.
         active_command_ids: Active command id set for MODIFY/ABORT checks.
-        max_order_rate: Optional max ENTER actions per second.
+        max_order_rate: Optional max ENTER actions/sec override.
+            Defaults to ``config.max_order_rate`` when not provided.
         now_fn: Optional clock override for deterministic tests.
 
     Returns:
@@ -59,8 +60,11 @@ def build_default_intake_hooks(
         ),
     ]
 
-    if max_order_rate is not None:
-        hooks.insert(0, make_order_rate_hook(max_order_rate, now_fn=now_fn))
+    resolved_max_order_rate = (
+        config.max_order_rate if max_order_rate is None else max_order_rate
+    )
+    if resolved_max_order_rate is not None:
+        hooks.insert(0, make_order_rate_hook(resolved_max_order_rate, now_fn=now_fn))
 
     return tuple(hooks)
 
--- a/nexus/core/validator/intake_stage.py
+++ b/nexus/core/validator/intake_stage.py
@@ -41,7 +41,7 @@ def build_default_intake_hooks(
     '''Build default intake hooks from operator configuration.
 
     Args:
-        config: Instance config carrying duplicate window settings.
+        config: Instance config carrying duplicate-window and order-rate settings.
         active_command_ids: Active command id set for MODIFY/ABORT checks.
         max_order_rate: Optional max ENTER actions/sec override.
             Defaults to ``config.max_order_rate`` when not provided.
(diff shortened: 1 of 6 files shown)


nexus/instance_config.py
```diff
@@ -38,6 +38,7 @@ class InstanceConfig:
     venue: str
     allocated_capital: Decimal
     duplicate_window_ms: int = 1000
+    max_order_rate: int | None = None
```
zero-bang, line 45: `max_order_rate` is a public field on a frozen dataclass but is not documented in the class-level `Args:` block. Every other field (`account_id`, `venue`, `allocated_capital`, `duplicate_window_ms`, `capital_pct`) has an entry there. Please add:

```
    max_order_rate: Optional cap on ENTER actions per second for intake
        rate-limiting. ``None`` disables the rate check.
```

nexus/core/validator/intake_stage.py
```diff
@@ -43,7 +43,8 @@ def build_default_intake_hooks(
     Args:
         config: Instance config carrying duplicate window settings.
```
zero-bang: `config` arg description is now stale — it says "duplicate window settings" but after this PR the config also drives `max_order_rate`. Update to something like: `Instance config carrying duplicate-window and order-rate settings.`

zero-bang (changes requested):
Logic and tests look solid. Two docstring gaps stand out — both in public API surfaces that the author's checklist explicitly covers.

nexus/instance_config.py
blahmonkey: Addressed in `be95116`. Added `max_order_rate` to `InstanceConfig` class docstring `Args:` with units and `None` disable semantics.

nexus/core/validator/intake_stage.py
blahmonkey: Addressed in `be95116`. Updated the `config` arg description in `build_default_intake_hooks` to mention both duplicate-window and order-rate settings.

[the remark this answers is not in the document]
blahmonkey: Addressed in `be95116`. `config` docstring now explicitly states duplicate-window + order-rate settings.

nexus/instance_config.py
```diff
@@ -30,6 +30,8 @@ class InstanceConfig:
             must not exceed this value.
         duplicate_window_ms: Duplicate-order detection window for intake
             checks, in milliseconds.
+        max_order_rate: Optional cap on ENTER actions per second for intake
+            rate-limiting. ``None`` disables the rate check.
```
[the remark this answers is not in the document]
blahmonkey: Addressed in `9dfaf22`. Updated `InstanceConfig.max_order_rate` docs to clarify this is a per-process intake limit, not a distributed/global rate limit across multiple processes/hosts.

zero-bang (approved):
Both items addressed. The `max_order_rate` entry is now in the `InstanceConfig` class `Args:` block with units, default semantics, and a clear per-process scope note. The `config` arg description in `build_default_intake_hooks` correctly reflects both duplicate-window and order-rate settings. Good to merge.

nexus/instance_config.py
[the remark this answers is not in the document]
blahmonkey: Addressed in `9dfaf22`. Clarified `max_order_rate` is per-process and explicitly not distributed/global across hosts/processes.

PR #24, opened 2026-03-30
feat: Phase 7.1 - Manifest loading and validation

### LINKED ISSUE #16: RFC-3001: Manager Sub-System
# RFC-3001 Manager

### Objective

The entire Vaquum Bitcoin-only research and trading system can be understood through four different aspects:

- Experiments
- Cohorts
- Manager
- Praxis (Trading sub-system + Account sub-system)

Experiments yield decoders, which come together in Cohorts.

Cohorts yield alpha signals, which come to the Manager (or third parties).

Manager yields trading decisions, which come to play in the Trading sub-system.

The Trading sub-system yields outcomes, which come to play in Experiments, Cohorts, Manager, and both sub-systems.

### Summary

Manager is the decision-making layer between signal generation and trade execution. It has two layers: a non-negotiable infrastructure layer (Core layer) that enforces risk, capital, health, and price validation rules, and a user-defined decision logic layer (Strategy layer) that determines when and how to trade.

### Is This Really Necessary?

Without Manager, nothing drives Trading. Cohorts produce signals; Trading executes orders; Manager is the bridge that turns one into the other. Required for end-to-end pipeline.

### Named Concepts

- **Manager instance**: Stateful process managing one trading account. Enforces risk/capital/health rules (Core layer) and runs user-defined decision logic (Strategy layer). One instance = one venue account = one API key = one trading pair = one quote asset. An instance can run multiple strategies internally
- **Core layer**: Non-negotiable infrastructure — risk validation, capital management, health policy, price validation, state tracking. Always enforced. Cannot be bypassed by Strategy layer
- **Strategy layer**: User-defined decision logic — when to enter, when to exit, how much, which execution mode. Pluggable. Runs inside Strategy Runner
- **Trade**: A strategic decision lifecycle — from entry to exit. Identified by `trade_id` (assigned by Manager instance). A trade may span multiple TradeCommands (e.g., entry command, then exit command)
- **predictor_fn**: A callable produced by either a Trainer (single decoder) or Cohort (multi-decoder aggregation). Both expose identical `predict()` contract. Manager instance treats them uniformly. A single strategy can consume multiple predictor_fns
- **Validator**: The Core layer validation pipeline. Receives actions from Strategy layer, validates against all risk rules, passes valid actions to Praxis Connector or returns denial to Strategy layer
- **Capital Controller**: Manages the instance's capital state — available capital, reservations, lifecycle state machine, drawdown tracking
- **Strategy Runner**: Hosts the Strategy layer, calls `predict()` on each predictor_fn at its configured `interval`, dispatches the resulting signals to the bound strategies, collects actions, submits to Validator
- **Reconciler**: Periodic state verification of Manager's internal state against Trading sub-system
- **Praxis Connector**: Sends TradeCommands to Praxis, receives execution results (TradeOutcomes, health, positions), updates instance state (positions, capital lifecycle, risk metrics), detects misfires
- **Decision Trail**: Append-only audit trail of every decision, gate check, and state change within the instance

### Out of Scope

- Market data sourcing, order book distribution, data infrastructure (Origo)
- Order execution, venue communication, position tracking, reconciliation against venue, kill switch (Praxis)
- Cross-instance aggregate risk (total drawdown/loss across multiple Manager instances) — platform reporting concern, not a single Manager instance concern

### Unresolved Questions

None.

---

#### Manager Instance

A Manager instance is the unit of deployment. One instance = one venue account = one API key = one trading pair = one quote asset. An instance can run multiple strategies internally, all sharing the same capital pool and venue connection.

- **Identity**: bound to one `account_id` registered with the Trading sub-system
- **Venue isolation**: one venue account per instance. Exchange enforces hard balance limit. No cross-instance capital sharing. Instances never share a venue account — no cross-instance capital coordination needed
- **Quote asset**: determined by the trading pair (BTC/USDT → USDT, BTC/EURC → EURC). All internal capital math in quote asset. One account, one quote asset. No mixed-denomination math, no currency conversion inside Manager. If a user wants to trade both BTC/USDT and BTC/EURC, that's two venue accounts with two Manager instances. Cross-instance P&L reporting in a single reference currency is a platform reporting concern, not a Manager concern
- **Strategies**: Strategy layer can run multiple strategies within one Manager instance. Capital contention between strategies handled by Capital Controller's atomic check-and-reserve mechanism
- **Config**: end-user configures Core layer (risk limits, health policy, etc.) and Strategy layer (strategy params) per instance. Platform operator optionally sets Gateway limits per client

#### Architecture

**Multi-Stage Risk Pipeline**

Two stages of validation before an order goes to Praxis:

1. **Core layer checks** (Validator): Manager risk gate. Non-bypassable. End-user configured. Capital, risk limits, price validation, health policy
2. **Gateway absolute limits**: final sanity check inside Validator before TradeCommand reaches Trading sub-system. Platform-operator configured (per-client), end-user cannot touch. These are "the building is on fire" limits

Gateway params (all optional — set per-client as needed):
- `GATEWAY_MAX_ORDER_NOTIONAL` — absolute dollar cap per single order
- `GATEWAY_MAX_ORDER_RATE` — absolute max orders per second
- `GATEWAY_MAX_POSITION` — absolute max position size
- `GATEWAY_MAX_DAILY_LOSS` — absolute max loss over rolling 24h
- `GATEWAY_MAX_CAPITAL_UTILIZATION` — absolute max capital deployed

These mirror critical Core layer limits with wider bounds (2-5x). Not tight constraints — catastrophe prevention only.

On Gateway rejection: enter reduce-only mode + alert platform ops. If Gateway fires, Core layer config is likely broken — new entries blocked, closes still allowed, platform ops investigates.

**Core / Strategy Interface**

Strategy layer decides what to do — enter, exit, modify, abort, or hold — and submits the action to Core layer with parameters (direction, size, execution mode, order type, deadline, execution params). The Validator validates the action against risk limits, capital availability, price sanity, and health status. If it passes, the action is handed to Praxis Connector which translates it into a venue-valid TradeCommand and sends to Trading sub-system. If it fails, the Validator returns a denial with reason.

The interface contract:
- Strategy layer receives: current positions, capital state, latest signals from predictor_fn(s), market data (price feed, order book depth, vol), health status
- Strategy layer submits: action object with parameters
- Validator checks: risk gate, capital gate, price validation, health policy
- Validator passes to: Praxis Connector (which emits TradeCommand to Trading sub-system) or returns denial with reason back to Strategy layer

Strategy layer has **no direct access** to the Trading sub-system API. It cannot submit orders, cancel orders, or modify orders directly. All actions flow through the Validator. Capability-based security — Strategy layer receives read-only access to state and a constrained action interface.

Risk denial is communicated back to Strategy layer so it can adjust strategy state, but Strategy layer cannot override the denial.

**Market Data (Origo)**

Manager subscribes to Origo's market data feed directly (order book, price, volume). Read-only subscription — no dedicated component needed. The Validator reads it for price checks (staleness, spread, deviation), Capital Controller reads it for mark-to-market (drawdown, unrealized P&L), Strategy layer reads it for decision-making.

#### Validator

The Core layer validation pipeline. Every action from Strategy layer passes through the Validator. If the action passes all checks, the Validator hands it to Praxis Connector for emission to Praxis. If it fails, the Validator returns a denial with reason to Strategy layer.

**Inbound validation (action from Strategy layer → TradeCommand or denial):**

The Validator runs checks in order. A failure at any stage returns a denial with reason to Strategy layer. Exit/close actions skip capital validation (exits release capital). Modify actions check only the delta if increasing. Cancel actions bypass `MAX_ORDER_RATE` — they are safety operations, not new risk (FIA: message throttles must never reject cancellations).

**Safety action bypass rule**: EXIT, ABORT, and CANCEL actions are risk-reducing operations. They must never be blocked by Gateway limits, HALTED operational mode, or health validation. These actions bypass stages 5 (health) and 6 (gateway) unconditionally. In HALTED mode, only EXIT/ABORT/CANCEL are permitted — all ENTER and MODIFY actions are rejected. In REDUCE_ONLY mode, ENTER is rejected, EXIT/ABORT/CANCEL/MODIFY are permitted.

Validation stages:

1. **Order intake checks**: `MAX_ORDER_RATE` (orders/sec from Strategy layer), `DUPLICATE_ORDER_WINDOW_MS` (same strategy_id/direction/size within window = likely bug — keyed by strategy_id so two strategies legitimately submitting identical orders don't collide). Reference-integrity: EXIT requires valid `trade_id` that exists and is open, EXIT `size` must be ≤ remaining position minus `pending_exit` for that trade_id (prevents double-exit), MODIFY/ABORT requires valid `command_id` in a modifiable/cancellable state, spot direction enforcement (BUY = entry, SELL = exit only). On violation: reject + alert (warning). These indicate Strategy layer bug
2. **Risk limit checks**: Per-strategy limits checked first (if configured in manifest `risk_limits`), then instance-level limits (always). Per-strategy: `max_position`, `max_notional`, `max_concurrent_trades`, `max_drawdown_pct`, `max_24h_loss`/`max_7d_loss`/`max_30d_loss`. Instance-level: `MAX_POSITION`, `MAX_NOTIONAL`, `MAX_CONCURRENT_TRADES`, `MAX_DRAWDOWN_PCT`, `MAX_24H_LOSS`/`MAX_7D_LOSS`/`MAX_30D_LOSS`, `MAX_POSITION_VELOCITY`. On violation: per-limit breach policy applies — per-strategy breach affects that strategy only, instance-level breach affects the entire instance (see Breach Policy and Per-Strategy Risk Limits below). `MIN_TRADE_NOTIONAL` — reject with reason, no alert (normal condition when strategy generates small sizing near boundaries)
3. **Price validation**: `BOOK_STALENESS_MAX_SECONDS`, `MAX_SPREAD_BPS`, `PRICE_DEVIATION_MAX_BPS` against `REFERENCE_PRICE_SOURCE`. On violation: `BOOK_STALENESS` — reject + system-level alert to platform ops AND end user (stale data = Origo may be down, infrastructure problem). `MAX_SPREAD` and `PRICE_DEVIATION` — reject with reason, no alert
4. **Capital validation**: delegates to Capital Controller for atomic check-and-reserve. On violation: reject + alert (warning). Capital rejection is not a risk breach — no escalation into breach policy
5. **Health validation**: checks health state consumed from Trading sub-system against three-threshold policy (warn/breach_action/halt) for latency, consecutive failures, failure rate, rate limit headroom, clock drift. On warn: alert. On breach: configurable `HEALTH_BREACH_ACTION` per metric (`reduce_only` or `throttle`). On halt: halt trading
6. **Gateway**: platform-operator absolute limits. EXIT/ABORT/CANCEL bypass this stage (safety action bypass rule). On violation for ENTER/MODIFY: reduce-only + alert platform ops

**Breach Policy**

Per-limit breach config following industry standard three-threshold model (FIA, NYSE Pillar, Ember OMS pattern). Each limit has its own breach thresholds and actions — no separate system-wide state machine.

For **continuous limits** (loss, drawdown, notional — can worsen without new orders due to unrealized P&L):
- `warn_pct` — alert threshold as fraction of limit (default 0.80). Alert only, trading continues
- `breach_action` — what happens at 100% of limit. Per-limit choice: `reduce_only` (block new entries, allow closes) or `reject` (reject the specific order)
- `hard_pct` — halt threshold as fraction of limit (default 1.20). Halts trading on this instance. Exists because unrealized losses can push past 100% without any new orders

Continuous limits are evaluated in two places: (1) on every action submission through the Validator, and (2) on a periodic mark-to-market tick driven by the Origo price feed (configurable via `CONTINUOUS_LIMIT_EVAL_INTERVAL_SECONDS`, default 10). The periodic tick re-evaluates drawdown and unrealized loss against `warn_pct`, `breach_action`, and `hard_pct` thresholds — without this, a position could blow past `hard_pct` between strategy submissions with no check catching it. The tick updates breach state and triggers mode transitions (reduce-only, halt) and alerts as needed, identical to the Validator path

For **discrete limits** (concurrent trades, position count, position velocity — can only breach via new order):
- `warn_pct` — alert threshold (default 0.80)
- `breach_action` — always `reject` (reject the order that would breach). No `hard_pct` needed — can't breach without a new order

**Per-Strategy Risk Limits**

Dual-layer risk model: per-strategy limits (if configured in manifest `risk_limits`) are checked first, then instance-level limits (always, as fallback).

Check order on every action:
1. **Per-strategy check**: the action's strategy_id metrics (drawdown, rolling losses, concurrent trades) vs that strategy's `risk_limits` from manifest. If breached → per-limit breach policy applies to **that strategy only** (reduce-only or halt that strategy, other strategies unaffected)
2. **Instance-level check**: aggregate instance metrics vs instance-level risk config. If breached → per-limit breach policy applies to the **entire instance**

Per-strategy limits are optional. If a strategy has no `risk_limits` in the manifest, only instance-level limits apply. Instance-level limits always run regardless — they are the safety net.

Available per-strategy risk params (same names as instance-level, scoped to one strategy's metrics):
- `max_drawdown_pct` — drawdown from this strategy's own HWM (computed from this strategy's positions and realized P&L)
- `max_24h_loss` / `max_7d_loss` / `max_30d_loss` — rolling loss from this strategy's trades only
- `max_concurrent_trades` — open positions for this strategy only
- `max_position` — position size for this strategy only
- `max_notional` — notional exposure for this strategy only

Per-strategy breach uses the same three-threshold model (warn/breach_action/halt). Breach policy config (`warn_pct`, `breach_action`, `hard_pct`) is inherited from instance-level breach policy — not separately configurable per strategy. Per-strategy `breach_action` defaults: `reduce_only` for continuous limits (halts that strategy's new entries), `reject` for discrete limits. Per-strategy `hard_pct` breach halts that strategy — **not** the instance.

This prevents a profitable strategy from masking a bleeding strategy's losses, and allows surgical intervention without halting the whole instance.

**Alert Routing**

Per-severity routing:
- `warning` — log, dashboard (approaching limit, informational)
- `error` — log, dashboard, webhook (soft breach, action taken)
- `critical` — log, dashboard, webhook, SMS (hard breach, halt triggered)

Webhook covers Slack/Discord/Telegram — user configures URL. SMS for critical alerts to key personnel (operator phone numbers).

#### Capital Controller

Manages the instance's capital state. The Validator delegates capital validation here (atomic check-and-reserve). The Praxis Connector delegates capital lifecycle transitions here (in_flight → working → position, or → released).

Two distinct concepts: `capital_pool` is a static budget ceiling set in the manifest — it defines the operational allocation and does not change from realized P&L (only changes via hot-reload). `equity` is the real-time financial state: cash + unrealized P&L at market. Equity is used for drawdown computation (HWM tracking). Per-strategy `strategy_realized_pnl` feeds into the effective budget calculation when `auto_compound` is enabled.

**Available capital computation:**

```
Available = capital_pool (from manifest)
          - Σ(position_notional)         # open positions at cost (BUY-side: quote locked in positions)
          - Σ(working_order_notional)     # resting BUY orders on venue (quote committed but not yet filled)
          - Σ(in_flight_order_notional)   # BUY orders sent but not yet acked (quote committed)
          - Σ(reservation_notional)       # TOCTOU locks not yet converted to orders (BUY-side only)
          - fee_reserve
```

All deductions are BUY-side (entry) only. SELL orders (exits) lock base asset (BTC) that is already accounted for in `position_notional` — they do not consume additional quote capital. SELL orders are not counted in `working_order_notional`, `in_flight_order_notional`, or `reservation_notional`. This is a two-asset market (BTC/quote): BUY locks quote, SELL locks base. The Available formula tracks quote commitment only.

`fee_rate` is consumed from Trading sub-system (Venue Adapter already fetches `GET /api/v3/account` for balance reconciliation — same response includes maker/taker commission fields). Fee estimates for SELL orders are deducted from the position's expected quote proceeds, not from Available — the quote returned from a SELL is `(fill_price × size) - fees`, and Available increases by that net amount on fill.

`fee_reserve` is a small buffer (default: `capital_pool × 0.01`) held back from Available to absorb fee rate fluctuations — if the venue changes fee tiers mid-session or a fill executes as taker instead of expected maker, actual fees may exceed per-order estimates. This is separate from the per-order `estimated_fees` included in each BUY reservation. Recalculated on manifest reload (capital_pool change) and fee schedule update from Trading sub-system.

FIA explicitly states working orders must be evaluated because they become positions if filled. Without this, two strategies can over-allocate from the same pool during volatile periods when multiple signals fire simultaneously.

**Capital validation (BUY/entry actions only):**

Check and reserve are one atomic operation — no race window between validation and reservation. This entire block applies to BUY (entry) actions only. SELL (exit) actions skip capital validation entirely — they release quote capital, they don't consume it. Modify actions check only the delta (new - old) if increasing.

```
acquire lock
  order_notional = size × price
  estimated_fees = order_notional × fee_rate
  base_budget = capital_pool × (strategy.capital_pct / 100)
  strategy_budget = base_budget + strategy_realized_pnl if auto_compound else base_budget

  check: order_notional / capital_pool ≤ MAX_ALLOCATION_PER_TRADE_PCT
  check: strategy_deployed + order_notional ≤ strategy_budget
  check: strategy_budget > 0  # compounding losses can exhaust budget
  check: available ≥ order_notional + estimated_fees
  check: (total_deployed + order_notional) / capital_pool ≤ MAX_CAPITAL_UTILIZATION_PCT

  if all pass → create reservation(order_notional + estimated_fees, TTL, strategy_id)
  if any fail → deny with reason, no reservation created
release lock
```

`strategy_deployed` = sum of positions + working BUY orders + in-flight BUY orders + reservations for this strategy_id. `total_deployed` = same across all strategies. `strategy_realized_pnl` = cumulative realized P&L for this strategy (positive for profits, negative for losses). Per-strategy budget check ensures no strategy exceeds its effective allocation. Instance-level utilization check ensures the overall ceiling holds.

For SELL actions: no capital validation, no reservation. Capital Controller instead records a `pending_exit` against the trade_id (size being exited). On fill, position decreases and Available increases by net proceeds `(fill_price × fill_qty - fees)`. On reject/cancel, `pending_exit` is cleared, position unchanged.

**Capital lifecycle state machine:**

BUY (entry) — locks quote capital:

```
reservation ──(order sent)──→ in_flight
in_flight ──(ack)──→ working
in_flight ──(reject)──→ released back to available
working ──(partial fill)──→ working decreases, position increases
working ──(full fill)──→ position at full notional
working ──(cancel)──→ released back to available
in_flight ──(timeout, no ack)──→ stays in_flight as "status unknown" until resolved
reservation ──(TTL expires)──→ released back to available
```

SELL (exit) — releases quote capital, locks base asset already in position:

```
position ──(exit order sent)──→ position unchanged, exit tracked as pending_exit
pending_exit ──(ack)──→ exit order resting on venue
pending_exit ──(reject)──→ position unchanged, pending_exit cleared
exit resting ──(partial fill)──→ position decreases, available increases by (fill_price × fill_qty - fees)
exit resting ──(full fill)──→ position closed, available increases by net proceeds
exit resting ──(cancel)──→ position unchanged, pending_exit cleared
pending_exit ──(timeout, no ack)──→ stays pending_exit as "status unknown" until resolved
```

SELL orders do not appear in `working_order_notional` or `in_flight_order_notional` — those track quote commitment (BUY-side only). SELL orders are tracked separately as `pending_exit` per trade_id to prevent double-exit (two strategies or two callbacks trying to exit the same position). Reference-integrity in Stage 1 validates EXIT size ≤ (remaining position - pending_exit size) for the trade_id.

Every transition is atomic — no intermediate state where capital is double-counted or unaccounted for.

**TOCTOU (Time-Of-Check-To-Time-Of-Use) race prevention:**

When multiple strategies (or multiple signal cadences) compete for capital from the same pool, a time-of-check-to-time-of-use race exists: both check available, both see $100K, both submit $80K, total $160K exceeds pool. The atomic check-and-reserve under lock prevents this.

Reservation TTL (default 30s, configurable via `RESERVATION_TTL_SECONDS`) prevents capital being locked indefinitely by abandoned requests. If order isn't submitted within TTL, reservation expires and capital returns to available pool. For single-strategy instances, contention is minimal — but the design supports multi-strategy from the start.

On violation: reject order, return denial with reason to Strategy layer, alert (warning). Capital rejection is not a risk breach — no escalation into breach policy. Reservation TTL expiry: capital returns to pool, alert (warning).

**Drawdown computation:**

Peak = total equity (cash + unrealized P&L at market). Lifetime high-water mark — never resets. No daily reset for crypto 24/7 markets; continuous rolling only. Daily snapshots for reporting, not risk logic.

`MAX_DRAWDOWN_PCT` checks total equity drawdown (realized + unrealized) — industry standard, the true risk state.

Track realized and unrealized separately for diagnostics:
- `total_drawdown` — total equity vs high_water_mark. This is what `MAX_DRAWDOWN_PCT` checks
- `realized_drawdown` — cash-only peak vs current cash. Locked-in losses from closed trades
- `unrealized_drawdown` — mark-to-market of open positions vs their cost basis

Mark-to-market: mid-price for P&L reporting, best bid for conservative risk limit checks (spot-only, long-or-zero — positions can only be long, so conservative valuation uses best bid, the price at which the position could actually be sold).

Diagnostic value: low realized + high unrealized = sitting on losing positions that won't close. High realized + low unrealized = took losses, remaining positions healthy. Different problems, different responses.

#### Praxis Connector

The interface between Manager and Praxis (Trading sub-system). Handles both outbound (sending TradeCommands) and inbound (receiving execution results).

**Outbound:**

Receives validated actions from Validator, translates into TradeCommands (attaching instance config such as `STP_MODE`), sends to Trading sub-system.

**Inbound (TradeOutcome from Trading → state updates):**

When execution results come back from Trading sub-system, Praxis Connector:
- Updates positions (entry fills grow, exit fills reduce)
- Delegates capital lifecycle transitions to Capital Controller (in_flight → working → position, or → released)
- Recalculates risk metrics at both instance-level and per-strategy level (rolling losses, drawdown vs HWM)
- Detects misfires (see below)
- Routes alerts per severity config

Also consumes from Trading sub-system:
- Health signals (latency, failures, clock drift, rate limit headroom) — fed to Validator for health policy checks
- Positions (pull interface for reconciliation) — fed to Reconciler
- Fee schedule (maker/taker commission per account) — fed to Capital Controller for fee reserve estimation

**Misfire Detection**

Always on. Known misfire types with production-standard responses:

| Misfire | Detection | Response | Alert | Halt? |
|---------|-----------|----------|-------|-------|
| Cancel-fill race | Fill received on order in PENDING_CANCEL state | Accept fill (terminal state wins), update position, notify Strategy layer to re-evaluate | INFO | NO — normal, happens routinely |
| Unexpected fill | Fill with no matching local trade_id | Accept fill (position accuracy wins), reduce-only, trigger reconciliation. Local state is wrong — you placed it but lost track (crash, dropped ack) | CRITICAL | YES — reduce-only until manual review |
| Duplicate outcome | Same trade_id terminal outcome received twice | Ignore duplicate (idempotency check), log. If conflicting terminal states: trigger reconciliation | WARNING (ERROR if repeated pattern) | NO |
| Phantom order | Order on venue with no local record | Cancel immediately, full halt, trigger reconciliation. State is fundamentally wrong | CRITICAL | YES — full halt, manual approval to resume |
| Stale fill | Fill timestamp significantly older than current time | Accept fill (position accuracy wins), flag for review. Investigate venue connectivity | WARNING (<1hr) / ERROR (>1hr) | NO — but trigger reconciliation if very stale |

#### Strategy Runner (Strategy layer)

Hosts the Strategy layer. Strategy Runner calls `predict()` on each registered predictor_fn at its configured `interval` (from instance config). The returned signal is dispatched to the strategy (or strategies) bound to that predictor_fn via the manifest. Strategy Runner collects the resulting actions and submits them to the Validator. Strategy layer decides what to do; Validator decides whether it's allowed.

**Signal Interpreter**

The core of each strategy. Consumes signals and context, produces actions (enter, exit, modify, abort, or hold). Each strategy is identified by `strategy_id` and bound to one or more predictor_fn(s).

Read-only access to all data within the Manager instance. What it uses is up to the strategy author.

*Signals (from predictor_fn):*
- ML model output from bound predictor_fn(s) — one per sensor or cohort
- Currently binary (CAN_ENTER=1, NO_PREDICTION=0). Future: confidence scores, directional strength, multi-class signals
- A strategy can consume multiple predictor_fns — sees an ensemble of signals

*Market state (from Origo):*
- Order book (bids, asks, depth at each level)
- Latest price (mid, last trade)
- Spread (bid-ask, in bps)
- Volume (rolling windows)
- Volatility (derived from price feed)

*Own state (from Instance State, scoped to this strategy_id):*
- Current positions (size, entry price, direction, unrealized P&L, position age)
- Capital available to this strategy
- Recent trade history (last N trades, win/loss, P&L)
- Rolling performance (win rate, consecutive wins/losses)

*Execution feedback (from Praxis Connector):*
- TradeOutcome results (FILLED, PARTIAL, REJECTED, EXPIRED, CANCELED + reason)
- Actual fill prices vs intended prices (slippage on recent fills)
- Fill rate on recent orders (are my limit orders getting filled?)

*System health (from Praxis Connector):*
- Venue latency, failure rate, rate limit headroom
- Operational mode (ACTIVE, REDUCE_ONLY, HALTED)

*Denial feedback (from Validator):*
- Why an action was rejected (which risk limit, capital insufficient, health, price validation)

**Strategy Logic**

The strategy author writes the logic that connects signal to action. This is the core of the Strategy layer — everything else (data access, validation, execution) is infrastructure. The Signal Interpreter receives data, the strategy logic decides what to do with it, and the action surface is how it expresses that decision.

Manager provides the data, the action interface, and the guardrails. The strategy author provides the decision logic: when to enter, when to exit, how much, which execution mode, at what price. This logic is entirely user-defined and pluggable — Manager is agnostic to what the logic does, as long as the resulting actions conform to the action interface.

**Events**

Event-driven invocation. The strategy author subscribes to the events they need — only subscribed events trigger the strategy's code. All read-only data (market state, positions, capital, health, operational mode) is available as context on every event, regardless of what triggered it.

| Event | Source | Schedule | When to subscribe |
|-------|--------|----------|-------------------|
| `on_signal` | Strategy Runner calls `predict()` on predictor_fn | Periodic, per predictor_fn `interval` (from instance config) | Almost always — this is the primary decision trigger |
| `on_outcome` | Praxis reports TradeOutcome | Async push, per command lifecycle | When the strategy needs to react to fills, rejections, or expiries |
| `on_timer` | Strategy-defined timer(s) | Periodic, strategy-configured interval | For position management between signals (trailing stops, TP checks, time-based exits) |
| `on_startup` | Strategy Runner, instance start | Once | Always — initialize strategy, load state |
| `on_shutdown` | Strategy Runner, instance stop | Once | Always — handle open positions, persist state |

A strategy can register multiple timers with different intervals (e.g., a 30-second timer for trailing stop checks and a 5-minute timer for position review). Each timer has a `timer_id` so the callback knows which timer fired.

A minimal strategy subscribes to `on_signal` only — react to predictor_fn signals, produce actions. A more sophisticated strategy adds `on_outcome` to react to execution results, and `on_timer` for position management between signals.

**Strategy Packaging**

One .py file per strategy, one manifest file for the instance.

The strategy file is pure decision logic — it implements event callbacks and nothing else. No infrastructure, no configuration, no predictor_fn wiring. The strategy author writes logic; the manifest declares how it runs.

The manifest configures all strategies for this Manager instance:

```yaml
capital_pool: 10000  # total capital in quote asset for this instance

strategies:
  - id: momentum_btc
    file: strategies/momentum.py
    capital_pct: 60
    auto_compound: true
    predictor_fns:
      - cohort_alpha_v3
      - sensor_volatility
    timers:
      - id: trailing_stop_check
        interval_seconds: 30
      - id: position_review
        interval_seconds: 300
    risk_limits:  # optional — per-strategy risk limits
      max_drawdown_pct: 0.15
      max_24h_loss: 500
      max_concurrent_trades: 3
    params:
      entry_threshold: 0.7
      sizing_method: fixed

  - id: mean_reversion_btc
    file: strategies/mean_reversion.py
    capital_pct: 40
    auto_compound: false
    predictor_fns:
      - cohort_beta_v2
    timers:
      - id: spread_check
        interval_seconds: 10
    risk_limits:
      max_drawdown_pct: 0.10
      max_24h_loss: 200
    params:
      lookback_periods: 20
      z_score_entry: 2.0
```

`capital_pool` is a static budget ceiling in quote asset for this instance — it does not change from realized P&L. Must not exceed `ALLOCATED_CAPITAL` (hard ceiling from instance config) or venue account balance — validated on manifest load and startup via Reconciler, rejected if over. `capital_pct` across all strategies must sum to 100. Validated on manifest load — rejected if not.

Each strategy's effective budget depends on `auto_compound` (default `false`):
- `auto_compound: false` — `strategy_budget = capital_pool × (capital_pct / 100)`. Fixed. Realized profits accumulate as excess cash but do not expand the budget
- `auto_compound: true` — `strategy_budget = (capital_pool × capital_pct / 100) + strategy_realized_pnl`. Realized profits grow the budget, realized losses shrink it. Uses realized P&L only — unrealized gains do not expand the budget (no sizing on paper profits). If losses consume the entire base budget, the strategy has zero effective budget and cannot enter new positions

Capital Controller enforces the effective budget per strategy_id on every check-and-reserve. The instance-level `MAX_CAPITAL_UTILIZATION_PCT` (default 0.80) still applies as an overall ceiling — even if a compounding strategy has grown its budget, the instance won't deploy more than 80% total across all strategies. `ALLOCATED_CAPITAL` remains the hard infrastructure cap regardless of compounding gains.

`risk_limits` per strategy are optional. When configured, the Validator checks per-strategy metrics against these limits first, then instance-level limits as fallback. A per-strategy breach halts only that strategy — other strategies and the instance continue. See Validator — Per-Strategy Risk Limits.

The manifest is hot-reloadable at runtime — see Hot Reload below.

The same .py file can appear in multiple manifest entries with different predictor_fns, capital_pct, and params — the logic is reusable, the configuration is per-instance.

Strategy Runner reads the manifest, loads strategy files, wires predictor_fn(s) to strategies, registers timers, and manages the event loop. The strategy file never touches any of this — it just receives events and returns actions.

**Hot Reload**

The manifest is the unit of hot reload. Strategy Runner detects manifest changes, diffs against current running state, and applies changes per tier. All changes are logged to Decision Trail with before/after values.

*Tier 1 — Immediate (apply on next tick, no disruption):*

| Change | Behavior |
|--------|----------|
| `capital_pool` increase | New capital available immediately |
| `capital_pool` decrease | If new pool < current deployment, affected strategies enter reduce-only until within budget. No forced liquidation |
| `capital_pct` rebalance | Must still sum to 100. If a strategy's new allocation < current deployment, that strategy enters reduce-only. If allocation increases, new headroom available immediately |
| `params` change | Next callback invocation sees new values. No state reset — strategy instance continues with existing state |
| `risk_limits` change | New limits effective on next validation check. Per-strategy metrics (HWM, rolling losses) are not reset — existing metrics checked against new thresholds immediately |
| `auto_compound` change | Effective on next capital validation. Toggling on: strategy_realized_pnl immediately factors into budget. Toggling off: budget reverts to static `capital_pool × capital_pct / 100`. If the strategy is currently deployed beyond the new (smaller) budget, it enters reduce-only |
| Timer added | New timer registered, starts firing |
| Timer removed | Timer deregistered, stops firing |
| Timer interval changed | Timer re-registered with new interval |

*Tier 2 — Lifecycle (controlled transition with position handling):*

| Change | Behavior |
|--------|----------|
| Strategy added | Rebalance capital_pct across all strategies → load .py file → instantiate class → call `on_startup` → register events. Existing strategies may enter reduce-only if their allocation shrank |
| Strategy removed | Call `on_shutdown` (author handles gracefully) → wait for returned actions to reach terminal state (timeout: `HOT_RELOAD_SHUTDOWN_TIMEOUT_SECONDS`, default 60) → if positions remain after timeout, Strategy Runner force-unwinds (ABORT in-flight + EXIT positions, same as exception handling) → release capital → deregister |
| `file` changed | Remove with full lifecycle (above) → add with fresh instantiation. No state carryover — `on_save()` from old instance is not passed to new code (new code may not understand old state format). New code starts clean via `on_startup` |
| `predictor_fns` added | New signals start flowing to strategy via `on_signal`. Strategy receives unfamiliar `predictor_fn_id` — author's responsibility to handle |
| `predictor_fns` removed | Signals from removed predictor_fn stop arriving. No automatic unwind — strategy continues with remaining predictor_fns. If the strategy depends on the removed predictor_fn, author handles gracefully in their logic |

Tier 2 changes are sequenced: remove operations complete (all positions terminal) before add operations begin. This prevents capital contention during transitions.

*Tier 3 — Forbidden (rejected on validation):*

| Change | Behavior |
|--------|----------|
| `id` changed | Rejected. Strategy ID is the key for trade_id mapping, position tracking, capital budgets, and Decision Trail correlation. Changing it orphans all state. To rename: remove old strategy (with unwind) + add new strategy |

*Manifest validation on hot reload:*

- `capital_pct` must sum to 100 — rejected if not
- `capital_pool` must not exceed `ALLOCATED_CAPITAL` or venue account balance — validated via Reconciler, rejected if over
- No duplicate strategy IDs
- All referenced .py files must exist and load without import errors
- All referenced `predictor_fn` IDs must be registered

On validation failure: reject entire manifest change, keep running with previous manifest, alert (ERROR) with specific validation failure reason.

**Interface Contract**

A strategy is a Python class with instance state. The strategy author writes the class — the Strategy Runner owns the execution loop. The strategy never calls `run()` or manages its own event loop. Strategy Runner receives events, dispatches to the appropriate callback, and forwards the returned actions to the Validator.

Instance state is entirely the strategy author's business. Manager doesn't define, constrain, or inspect strategy state — it just provides the persistence mechanism. The author uses instance variables for signal history, indicator values, trade context, cooldown tracking, or anything else their logic needs.

```python
class MyStrategy:
    def __init__(self):
        self.signal_history = []
        self.consecutive_losses = 0
        self.last_entry_reason = None

    def on_signal(self, signal, params, context):
        self.signal_history.append(signal)
        if len(self.signal_history) < params.min_signals:
            return []  # not ready yet — self-gating warmup
        # ... author's decision logic ...
```

Strategy Runner calls `on_save() → bytes` for state persistence (shutdown, periodic checkpoints) and `on_load(bytes)` for state restoration (startup, crash recovery). The author serializes whatever they want — infrastructure doesn't care what's inside.

**Crash recovery replay**: After `on_load()` restores the strategy's last checkpoint, Strategy Runner replays all strategy events logged in the WAL since that checkpoint — calling the strategy's callbacks (on_signal, on_outcome, on_timer) with the original payloads in order. Actions returned during replay are **discarded** (the orders were already submitted pre-crash — Manager-level state is already correct from WAL replay and Reconciler). The purpose is solely to bring the strategy's internal state (signal history, indicator values, trade context) up to date with what actually happened. After replay completes, the strategy resumes live event processing with consistent state.

Event callbacks:

```
on_startup(params, context) → list[Action]
on_signal(signal, params, context) → list[Action]
on_outcome(outcome, params, context) → list[Action]
on_timer(timer_id, params, context) → list[Action]
on_shutdown(params, context) → list[Action]
on_save() → bytes
on_load(bytes) → None
```

- `params`: the strategy's configuration from the manifest (e.g., `params.entry_threshold`, `params.sizing_method`)
- `signal`: the predictor_fn output (signal value, predictor_fn_id, timestamp)
- `outcome`: the TradeOutcome from Praxis (trade_id, command_id, status, fill data, reason). Scoped to this strategy's trades only — a strategy has no visibility into sibling strategies
- `timer_id`: which timer fired
- `context`: all read-only data — positions, capital, market state, health, execution feedback, operational mode
- Returns: zero or more Action objects conforming to the action surface below. HOLD is implicit when the list is empty

Callbacks are serialized — one at a time per strategy. No concurrent callbacks, no state corruption risk.

`on_startup` returns actions for strategies that need to act on initialization (e.g., place standing orders). `on_shutdown` returns actions for position handling (e.g., ABORT all in-flight commands, EXIT all positions).

No formal warmup phase. Strategies that need to accumulate data before trading (e.g., N signals for a rolling average) self-gate by returning empty actions until ready. Simpler and more flexible than a framework-level warmup lifecycle.

**Exception Handling**

If a strategy's callback throws an exception, Strategy Runner treats it as a critical failure for that strategy:

1. ABORT all in-flight commands for this strategy_id
2. EXIT all open positions (market order, short deadline) — Strategy Runner performs the unwind, not the broken strategy code
3. Halt this strategy — no further callbacks
4. Alert (CRITICAL) with exception details and stack trace
5. Other strategies in the instance continue unaffected

If the unwind itself fails (Praxis rejects the exit), escalate to instance-level alert — human intervention needed.

**Identifiers**

`trade_id` is assigned by Core layer when an ENTER action passes validation. The strategy receives it back via `on_outcome` and uses it for subsequent EXIT/MODIFY/ABORT actions. `command_id` is returned by Praxis on TradeCommand submission and delivered to the strategy via `on_outcome`. `symbol` is implicit from instance configuration (one instance = one pair).

**Action Types**

| Action | What it means | Maps to (Praxis) |
|--------|--------------|-------------------|
| ENTER | Open new position or add to existing trade | TradeCommand (BUY) with new or existing trade_id |
| EXIT | Close/reduce a position | TradeCommand (SELL) on existing trade_id |
| MODIFY | Change params on an in-flight command | TradeModify (SingleShot, Bracket, Time DCA only; others require ABORT + new ENTER) |
| ABORT | Cancel an in-flight command | TradeAbort |
| HOLD | Do nothing — explicit decision not to act | No submission |

**Action Parameters (ENTER / EXIT)**

| Parameter | Description | Required |
|-----------|-------------|----------|
| `trade_id` | Existing trade_id for EXIT (received via `on_outcome` after ENTER). Not set on ENTER — assigned by Core layer on validation | EXIT only |
| `direction` | BUY or SELL. Spot-only: BUY = entry, SELL = exit | Yes |
| `size` | Quantity in base asset (BTC) | Yes |
| `execution_mode` | SingleShot, Bracket, TWAP, Scheduled VWAP, Iceberg, Time DCA, Ladder DCA | Yes |
| `order_type` | Market, Limit, Limit IOC, Stop, Stop-Limit, TP, TP-Limit, OCO. Constrained by mode | Yes |
| `deadline` | Timeout in seconds. Praxis reports EXPIRED if exceeded | Yes |
| `execution_params` | Mode-specific params (price, duration, slice config — see mode table below) | Yes |
| `maker_preference` | MAKER_ONLY / MAKER_PREFERRED / NO_PREFERENCE (default: NO_PREFERENCE). MAKER_ONLY maps to post_only/LIMIT_MAKER on venue — guarantees maker status and fee rebate, rejects if order would cross | No |
| `reference_price` | Supplementary reference from strategy. Praxis uses walk-the-book for primary slippage estimation | No |

**Action Parameters (MODIFY)**

| Parameter | Description | Required |
|-----------|-------------|----------|
| `command_id` | Which in-flight command to modify (returned by Praxis on original submission) | Yes |
| `new_params` | Updated execution_params (replaces corresponding fields) | Yes |
| `deadline` | Updated timeout | No |

**Action Parameters (ABORT)**

| Parameter | Description | Required |
|-----------|-------------|----------|
| `command_id` | Which in-flight command to cancel | Yes |

**Mode ↔ Order Type Constraints**

Defined by Praxis. Strategy must respect these when constructing actions.

| Mode | Allowed order_type | execution_params |
|------|-------------------|-----------------|
| SingleShot | Market, Limit, Limit IOC, Stop, Stop-Limit, TP, TP-Limit, OCO | price, stop_price, stop_limit_price (as needed by order type) |
| Bracket | Entry: Market, Limit, Limit IOC, Stop, Stop-Limit. TP/SL auto-placed | entry_price, tp_price, sl_price, stop_price |
| TWAP | Market, Limit, Limit IOC | total_duration, slice_count or slice_interval, price |
| Scheduled VWAP | Market, Limit, Limit IOC | slices: [{qty, time}] — pre-computed by strategy, price |
| Iceberg | Limit | visible_slice_size, replenishment_delay, price |
| Time DCA | Market, Limit, Limit IOC | interval, amount_per_interval, total_count or total_duration, price, missed_interval_policy |
| Ladder DCA | Limit, Stop-Limit | levels: [{price, qty}] or [{price, qty, stop_price}] |

**Modify Support per Mode**

| Mode | Modifiable? | How |
|------|------------|-----|
| SingleShot | Yes (resting limit only) | Cancel-replace via Venue Adapter |
| Bracket | Yes (resting legs only) | Cancel-replace TP/SL legs |
| Time DCA | Yes (future iterations) | Internal schedule update |
| TWAP, Scheduled VWAP, Iceberg, Ladder DCA | No | Strategy must ABORT + new ENTER for remaining qty |

**Denial / Rejection / Expiry Handling**

When an action is denied or an execution fails, the strategy receives feedback and decides what to do next. Re-entry logic is entirely strategy-authored.

| Event | Source | What strategy receives | Strategy decides |
|-------|--------|----------------------|-----------------|
| Risk denial | Validator | Rejection reason (which limit, capital, health, price) | Retry with different params, reduce size, or give up |
| REJECTED | Praxis | TradeOutcome with reason (venue rejection) | Resubmit with different order_type or params, or give up |
| EXPIRED | Praxis | TradeOutcome with fill summary (partial fills if any) | Resubmit remainder with different params, or accept partial |
| PARTIAL + slice_failed | Praxis | Non-terminal TradeOutcome with failure reason | MODIFY, ABORT, or wait |
| CANCELED | Praxis | TradeOutcome with cumulative fill data | Resubmit or accept |

#### Reconciler

Periodic state verification of Manager's internal state against Trading sub-system. Configured via `RECONCILIATION_INTERVAL_SECONDS` (default 60).

What gets compared:
1. Manager's per-strategy gross positions vs exchange net position (invariant: `Σ(strategy_positions) = exchange_net_position`)
2. Manager's capital state (pool, deployed, available) vs Trading's Account balance
3. Manager's working/in_flight orders vs Trading's open orders on venue

**In-flight-aware comparison**: Before escalating, Reconciler checks whether a mismatch is explainable by orders currently in `working` or `in_flight` state. If a position delta matches a pending order that could have filled (size and direction consistent), the mismatch is downgraded to WARNING with tag "likely in-flight fill" — not CRITICAL. Only unexplainable deltas (no matching pending order, or delta in wrong direction) escalate to CRITICAL. This prevents false alerts from fills that are in transit between Praxis and Manager's processing pipeline.

On unexplainable mismatch: alert (CRITICAL), enter reduce-only mode, log full state diff (Manager belief vs Trading reality). Requires manual review to resume — state divergence means something is fundamentally wrong. Minor rounding differences (< fee tolerance) logged as WARNING, not escalated.

#### Decision Trail

Manager-level decision journal — every Strategy layer decision, every Core layer gate check, every action taken. "Why did it enter this trade? What signal? What was the capital state? What risk checks passed?"

- Decision Trail covers decision audit only ("why did we trade?"). Execution audit ("how was the order handled?") lives in Praxis — MiFID II explicitly requires separate documentation for each
- Event-sourced append-only log
- Correlation IDs linking the full decision chain: signal_id → risk_check_id → action_id → trade_command_id → trade_outcome_id
- Hash-chaining for tamper evidence
- Storage tier strategy needed for production: hot (recent, fast query) → warm (weeks) → cold (archive)
- Config changes (Core layer parameters) logged as first-class events with before/after values. Treat parameter changes like production releases: logged, reviewed, reversible
- Manifest hot-reload changes logged as first-class events: capital_pool changes, capital_pct rebalances, strategy additions/removals, file changes, predictor_fn changes, param changes, timer changes — all with before/after values and tier classification

#### Instance State

Everything a Manager instance tracks at runtime. Managed across Praxis Connector, Capital Controller, Validator, and Reconciler.

**Capital (Capital Controller):**
- `capital_pool` — static budget ceiling for this instance, denominated in quote asset (from manifest; must be ≤ ALLOCATED_CAPITAL and ≤ venue balance)
- `position_notional` — Σ open positions at cost
- `working_order_notional` — Σ resting BUY orders on venue (quote committed, not yet filled). SELL orders excluded — they lock base, not quote
- `in_flight_order_notional` — Σ BUY orders sent but not yet acked (quote committed). SELL orders excluded
- `pending_exit` — per trade_id, size of SELL orders in-flight or resting. Prevents double-exit. Not subtracted from Available (no quote locked)
- `fee_reserve` — reserved for transaction costs
- `capital_reservations` — active TOCTOU locks with TTLs (BUY-side only)
- `available` — derived: pool - positions - working - in_flight - fee_reserve - reservations (all BUY-side deductions)
- `equity` — derived: cash (capital_pool + Σ realized_pnl - Σ deployed) + Σ unrealized_pnl. Used for drawdown computation. Distinct from `capital_pool` (static budget ceiling)

**Risk tracking (Praxis Connector updates, Validator checks):**

Instance-level:
- `high_water_mark` — lifetime peak total equity (for drawdown)
- `rolling_loss_24h` — rolling 24-hour realized loss
- `rolling_loss_7d` — rolling 7-day realized loss (optional)
- `rolling_loss_30d` — rolling 30-day realized loss (optional)
- `realized_pnl` — cumulative realized P&L

Per-strategy (keyed by strategy_id, computed from trade_id → strategy_id mapping):
- `strategy_high_water_mark` — lifetime peak equity for this strategy's positions + realized P&L
- `strategy_rolling_loss_24h` — rolling 24-hour realized loss from this strategy's trades
- `strategy_rolling_loss_7d` — rolling 7-day realized loss (optional, if per-strategy limit configured)
- `strategy_rolling_loss_30d` — rolling 30-day realized loss (optional, if per-strategy limit configured)
- `strategy_realized_pnl` — cumulative realized P&L for this strategy

**Per-limit breach state (Validator manages):**
- Instance-level: which limits are at warn / breach_action / halt threshold
- Per-strategy: which strategy-level limits are at warn / breach_action / halt threshold (independent per strategy)

**Health state (consumed from Trading sub-system):**
- Latest latency p99, consecutive failures, failure rate, rate limit headroom, clock drift

**Positions:**
- Open trades with entry price, size, direction, trade_id, strategy_id
- Per-trade unrealized P&L (mark-to-market from Origo price feed)
- Per-strategy gross positions tracked internally (strategy A: 0.5 BTC, strategy B: 0.3 BTC). Exchange only knows the net (0.8 BTC). All long-or-zero in spot — no short positions to net against
- Invariant: `Σ(strategy_positions) = exchange_net_position` — validated on every reconciliation cycle
- P&L attribution per strategy via trade_id → strategy_id mapping (Manager assigns trade_id per strategy, TradeOutcomes route back referencing trade_id)

**Gateway state:**
- Platform-operator limits (if set for this client)

**Operational mode:**
- Instance-level mode: `ACTIVE` (normal trading), `REDUCE_ONLY` (new entries blocked, closes allowed), or `HALTED` (all trading stopped, manual approval to resume)
- Per-strategy mode: each strategy independently tracks `ACTIVE`, `REDUCE_ONLY`, or `HALTED`. A strategy can be reduce-only (capital_pct shrank below deployment, hot-reload transition, or per-strategy risk limit breach_action) or halted (exception thrown, or per-strategy risk limit hard_pct breach) while the instance and other strategies remain active
- Instance-level mode overrides per-strategy mode — if the instance is HALTED, all strategies are halted regardless of their individual mode
- Trigger tracking: what caused each mode transition (breach limit, Gateway rejection, misfire type, reconciliation mismatch, health halt, exception, hot-reload capital rebalance)

All of the above is persisted (see Persistence) and recoverable from WAL replay + Trading sub-system reconciliation on crash restart.

#### Configuration

**Instance identity (required, no defaults):**

| Config | Description |
| ------ | ----------- |
| `ACCOUNT_ID` | Unique identifier for this Manager instance's trading account |
| `VENUE` | Which venue to trade on (e.g., `binance_spot`) |
| `API_KEY` | Venue API key for this account |
| `API_SECRET` | Venue API secret for this account |
| `ALLOCATED_CAPITAL` | Hard ceiling on capital this Manager instance can use, in quote asset. The manifest's `capital_pool` must be ≤ this value — rejected on load if over. `ALLOCATED_CAPITAL` is the infrastructure limit (set once, rarely changed); `capital_pool` is the operational allocation (set in manifest, hot-reloadable) |

**Signal source:**

Defines which predictor_fn(s) this instance can subscribe to. The mapping of predictor_fn → strategy is declared in the strategy manifest (`strategies[].predictor_fns`). A strategy can consume multiple predictor_fns.

| Config | Description |
| ------ | ----------- |
| `decoder_type` | Which decoder to use |
| `is_cohort` | Boolean — whether the predictor is a Cohort or single Trainer |
| `permutation_ids` | Array of permutation IDs to subscribe to (single element if not cohort) |
| `interval` | How often to invoke predict() |

**Risk limits:**

| Config | Description |
| ------ | ----------- |
| `MAX_POSITION` | Max position size in base asset |
| `MAX_NOTIONAL` | Max notional exposure in quote asset |
| `MAX_CONCURRENT_TRADES` | Max number of open positions at once. Without this, Strategy layer can open unlimited positions |
| `MAX_DRAWDOWN_PCT` | Max drawdown percentage from peak |
| `MAX_24H_LOSS` | Hard loss limit in quote asset over a rolling 24-hour window |
| `MAX_7D_LOSS` | Optional loss limit over a rolling 7-day window |
| `MAX_30D_LOSS` | Optional loss limit over a rolling 30-day window |
| `MAX_POSITION_VELOCITY` | Max position change per rolling window (e.g., per minute). Catches runaway algorithms that accumulate positions within individual limits but at abnormal rate. Knight Capital lesson |
| `MAX_ORDER_RATE` | Max orders per second from Strategy layer to Manager. Manager-side intake valve — catches runaway strategy bugs. Cancel and close actions bypass this throttle (FIA: message throttles must never reject cancellations). Venue-side rate limit budget allocation is a Trading sub-system concern |
| `DUPLICATE_ORDER_WINDOW_MS` | Time window for detecting duplicate order submissions. Same strategy_id/direction/size within window = likely bug. Keyed by strategy_id — two strategies can legitimately submit identical order shapes. Catches strategy retry loops |
| `MIN_TRADE_NOTIONAL` | Absolute floor on trade size in quote asset. Binance's MIN_NOTIONAL (~$5) is the venue minimum — technically valid but economically absurd at small sizes. This is the economic minimum: below this, round-trip fees (~0.20% at default tier) + typical spread cost dominate the trade |
| `STP_MODE` | Self-trade prevention policy. Industry-standard across centralized venues (Binance, Coinbase, Kraken, OKX, Bybit). Enum: `cancel_maker`, `cancel_taker` (default), `cancel_both`. Mapping to venue-specific params lives in Trading sub-system's Venue Adapter |

**Price validation:**

| Config | Description |
| ------ | ----------- |
| `BOOK_STALENESS_MAX_SECONDS` | If order book data from Origo is older than this, refuse to trade. Stale data → bad fills |
| `MAX_SPREAD_BPS` | Max bid-ask spread before refusing to trade. Wide spread = low liquidity = bad fills |
| `REFERENCE_PRICE_SOURCE` | Baseline for deviation check: `mid` (best bid + best ask / 2) or `last_trade` |
| `PRICE_DEVIATION_MAX_BPS` | Max deviation from reference price before rejection. Catches Strategy layer bugs producing nonsensical prices |

**Health policy:**

Same three-threshold model as Breach Policy (warn → breach_action → halt). Manager consumes health signals from Trading sub-system — it doesn't manage connections or parse headers directly.

| Config | Description |
| ------ | ----------- |
| `HEALTH_LATENCY_WARN_MS` | Ack latency p99 alert threshold (default 200ms) |
| `HEALTH_LATENCY_BREACH_MS` | Ack latency breach action threshold (default 500ms) |
| `HEALTH_LATENCY_HALT_MS` | Ack latency halt threshold (default 1000ms) |
| `HEALTH_FAILURE_WARN` | Consecutive failure alert threshold (default 3) |
| `HEALTH_FAILURE_BREACH` | Consecutive failure breach action threshold (default 5) |
| `HEALTH_FAILURE_HALT` | Consecutive failure halt threshold (default 10) |
| `HEALTH_FAILURE_RATE_WINDOW_SECONDS` | Rolling window for failure rate calculation (default 60s) |
| `HEALTH_FAILURE_RATE_WARN` | Failure rate alert threshold (default 0.10) |
| `HEALTH_FAILURE_RATE_BREACH` | Failure rate breach action threshold (default 0.20) |
| `HEALTH_FAILURE_RATE_HALT` | Failure rate halt threshold (default 0.40) |
| `HEALTH_HEADROOM_WARN` | Rate limit headroom alert threshold (default 0.70) |
| `HEALTH_HEADROOM_BREACH` | Rate limit headroom breach action threshold (default 0.85) |
| `HEALTH_HEADROOM_HALT` | Rate limit headroom halt threshold (default 0.90) |
| `HEALTH_CLOCK_DRIFT_MAX_MS` | Max allowed clock drift from exchange before halting. Binance hard-rejects at 1000ms, halt before that (default 500ms) |
| `HEALTH_ORDER_ACK_TIMEOUT_MS` | How long to wait for order ack before treating as "status unknown" (default 5000ms) |
| `RECONCILIATION_INTERVAL_SECONDS` | How often Reconciler syncs internal state against Trading sub-system (default 60). See Reconciler for comparison details and mismatch response |
| `HEALTH_BREACH_ACTION` | Default breach action for all health metrics: `reduce_only` or `throttle` (default `reduce_only`). Per-metric override possible if needed |

**Capital guardrails:**

Manager sets ceilings, Strategy layer sizes within them. Position sizing method (fixed, risk-based, volatility-adjusted, Kelly) is a Strategy layer decision — Manager just validates the number against limits.

| Config | Description |
| ------ | ----------- |
| `MAX_ALLOCATION_PER_TRADE_PCT` | Max position value per trade as fraction of total capital (default 0.15) |
| `MAX_CAPITAL_UTILIZATION_PCT` | Max total capital deployed across all positions simultaneously (default 0.80). Without this, Strategy layer can open multiple positions that individually pass per-trade limits but collectively overcommit capital |
| `RESERVATION_TTL_SECONDS` | Capital reservation timeout for TOCTOU prevention (default 30). When Strategy layer requests capital, it's reserved. If order isn't submitted within TTL, reservation expires and capital returns to available pool |

**Breach policy:**

Per-limit config. See Validator — Breach Policy for the model.

| Config | Description |
| ------ | ----------- |
| `warn_pct` | Per-limit alert threshold as fraction of limit (default 0.80) |
| `breach_action` | Per-limit action at 100%: `reduce_only` or `reject` |
| `hard_pct` | Per-limit halt threshold (default 1.20, continuous limits only) |
| `CONTINUOUS_LIMIT_EVAL_INTERVAL_SECONDS` | How often to re-evaluate continuous limits (drawdown, unrealized loss) via mark-to-market from Origo price feed (default 10). Catches breaches from price movement between strategy submissions |

**Alert routing:**

| Config | Description |
| ------ | ----------- |
| `ALERT_DESTINATIONS` | Per-severity routing: warning (log, dashboard), error (log, dashboard, webhook), critical (log, dashboard, webhook, SMS). Webhook covers Slack/Discord/Telegram — user configures URL. SMS for critical alerts to key personnel |

**Infrastructure:**

| Config | Description |
| ------ | ----------- |
| `MANAGER_STATE_DB_PATH` | Where to store Manager instance state |
| `STRATEGY_MANIFEST_PATH` | Path to the strategy manifest file. Strategy Runner watches this for hot-reload changes |
| `STATE_CHECKPOINT_INTERVAL_SECONDS` | How often Strategy Runner calls `on_save()` on each strategy for periodic state persistence between shutdowns (default 300) |
| `HOT_RELOAD_SHUTDOWN_TIMEOUT_SECONDS` | When a strategy is removed via hot reload, how long to wait for `on_shutdown` actions to reach terminal state before force-unwinding (default 60) |
| `SHUTDOWN_TIMEOUT_SECONDS` | When the instance shuts down, how long to wait for all commands to reach terminal state before force-aborting remaining in-flight orders (default 120). Prevents shutdown from hanging indefinitely on stuck orders |

#### Startup / Shutdown

**Startup:**

Load persisted state → replay WAL from last snapshot → register with Trading sub-system → reconcile capital state against Trading's Positions → load and validate strategy manifest (capital_pool ≤ ALLOCATED_CAPITAL, capital_pct sums to 100, .py files load, predictor_fn IDs registered) → instantiate strategy classes → call `on_load(bytes)` for state restoration → replay strategy events from WAL since last checkpoint (actions discarded — state catch-up only) → receive predictor_fn(s) (one per sensor or cohort) → register timers → enter ACTIVE or REDUCE_ONLY mode based on health → call `on_startup` on each strategy (actions go through Validator with operational mode already set).

**Shutdown:**

Stop accepting new signals → call `on_shutdown` on each strategy (returns actions for position handling — close all, hold, or let existing commands reach terminal state) → submit returned actions through Validator (exit/abort actions bypass health and operational mode checks — shutdown must be able to close positions regardless of instance state) → wait for all commands to reach terminal state (timeout: `SHUTDOWN_TIMEOUT_SECONDS`, default 120 — if commands are still in-flight after timeout, force-abort remaining orders and alert CRITICAL) → call `on_save()` on each strategy for state persistence → persist Instance State → deregister from Trading sub-system.

**Crash-only design:**

Startup and crash recovery share the same code path. No special recovery mode.

#### Persistence

Two logs, different purposes:
- **WAL (Write-Ahead Log)** — recovery source. Compact log of state mutations. Every state change is written to WAL before being applied in-memory. Truncated after periodic snapshots. Optimized for fast crash recovery. Also logs lightweight strategy event records (see below)
- **Decision Trail** — audit source. Rich log with full context (signal values, risk check results, capital state, correlation IDs, hash-chaining). Never truncated — stored in tiers (hot → warm → cold). Optimized for "why did we trade?" Not used for recovery

**Strategy event logging in WAL**: Every event delivered to a strategy callback (on_signal, on_outcome, on_timer) is logged to the WAL as a lightweight record: `(strategy_id, event_type, payload, timestamp)`. These records enable strategy state replay on crash recovery — see Strategy Runner for replay behavior. The records are truncated along with the rest of the WAL after each `on_save()` checkpoint for that strategy.

Persisted state: everything in Instance State. All fields are recoverable from WAL replay + Trading sub-system reconciliation, but snapshots avoid full replay on every restart.

Recovery after crash: load latest snapshot → replay WAL from snapshot point to reconstruct Manager-level state → reconcile capital state against Trading's Positions → release expired reservations → re-derive rolling loss counters from WAL events → restore strategy state via `on_load()` then replay strategy events from WAL since last checkpoint (see Strategy Runner). Decision Trail is not part of the recovery path — it is append-only audit

### Work Phases

Each checkbox belongs to exactly one work package phase: `[MMVP]`, `[MVP]`, or `[Later]`. Estimate: ~0.5d per checkbox. Total: 65 checkboxes, ~32.5d (14d MMVP, 16d MVP, 2.5d Later).

#### Phase 0 — Setup (~1d: 1 MMVP)

- [ ] 0.1 Finalize RFC `[MMVP]`
- [ ] 0.2 Project scaffolding: module structure, config loading, logging `[MMVP]`

#### Phase 1 — Instance State & Persistence (~1.5d: 1.5 MMVP)

- [ ] 1.1 Domain dataclasses: Instance State (capital, risk tracking, positions, operational mode) `[MMVP]`
- [ ] 1.2 WAL-based persistence: write-ahead log for state changes `[MMVP]`
- [ ] 1.3 State recovery: load persisted state, replay WAL for loss counters `[MMVP]`

#### Phase 2 — Capital Controller (~4d: 3.5 MMVP, 0.5 MVP)

- [ ] 2.1 Available capital computation (pool - positions - working - in_flight - reservations - fee_reserve) `[MMVP]`
- [ ] 2.2 Atomic check-and-reserve under lock with strategy_budget enforcement `[MMVP]`
- [ ] 2.3 Capital lifecycle state machine (reservation → in_flight → working → position, with all release paths) `[MMVP]`
- [ ] 2.4 Reservation TTL expiry `[MMVP]`
- [ ] 2.5 Drawdown computation: instance-level HWM, total/realized/unrealized `[MMVP]`
- [ ] 2.6 Per-strategy capital tracking (strategy_deployed, strategy_budget from capital_pct) `[MMVP]`
- [ ] 2.7 SELL lifecycle: pending_exit tracking per trade_id, double-exit prevention `[MMVP]`
- [ ] 2.8 Auto-compound: per-strategy budget adjustment from realized P&L when enabled `[MVP]`

#### Phase 3 — Validator (~5d: 2.5 MMVP, 2.5 MVP)

- [ ] 3.1 Validation pipeline: 6-stage ordered checks (intake → risk → price → capital → health → gateway) `[MMVP]`
- [ ] 3.2 Order intake checks: MAX_ORDER_RATE, DUPLICATE_ORDER_WINDOW_MS `[MMVP]`
- [ ] 3.3 Instance-level risk limit checks with breach policy (warn/breach_action/halt) `[MMVP]`
- [ ] 3.4 Price validation: staleness, spread, deviation against Origo feed `[MMVP]`
- [ ] 3.5 Health validation: three-threshold policy for latency, failures, rate limit headroom, clock drift `[MVP]`
- [ ] 3.6 Gateway absolute limits (platform-operator configured) `[MVP]`
- [ ] 3.7 Per-strategy risk limit checks (dual-layer: per-strategy first, instance fallback) `[MVP]`
- [ ] 3.8 Per-strategy breach state tracking and operational mode transitions `[MVP]`
- [ ] 3.9 Reference-integrity checks: trade_id exists, EXIT ≤ position - pending_exit, MODIFY/ABORT valid command_id `[MMVP]`
- [ ] 3.10 Continuous limit periodic eval (CONTINUOUS_LIMIT_EVAL_INTERVAL_SECONDS) for drawdown/loss breaches between submissions `[MVP]`

#### Phase 4 — Praxis Connector (~2.5d: 1.5 MMVP, 1 MVP)

- [ ] 4.1 Outbound: translate validated actions → TradeCommands, attach STP_MODE, send to Trading sub-system `[MMVP]`
- [ ] 4.2 Inbound: receive TradeOutcomes, update positions, delegate capital lifecycle to Capital Controller `[MMVP]`
- [ ] 4.3 Inbound: recalculate risk metrics (instance-level + per-strategy rolling losses, drawdown vs HWM) `[MMVP]`
- [ ] 4.4 Consume health signals, positions (pull), fee schedule from Trading sub-system `[MVP]`
- [ ] 4.5 Misfire detection: cancel-fill race, unexpected fill, duplicate outcome, phantom order, stale fill `[MVP]`

#### Phase 5 — Decision Trail (~2d: 1 MVP, 1 Later)

- [ ] 5.1 Append-only event log: signal → risk_check → action → trade_command → trade_outcome correlation `[MVP]`
- [ ] 5.2 Config change logging (Core layer params + manifest hot-reload) with before/after `[MVP]`
- [ ] 5.3 Hash-chaining for tamper evidence `[Later]`
- [ ] 5.4 Storage tiers: hot → warm → cold `[Later]`

#### Phase 6 — Reconciler (~2.5d: 2.5 MVP)

- [ ] 6.1 Periodic reconciliation: per-strategy positions vs exchange net position `[MVP]`
- [ ] 6.2 Capital state reconciliation: pool/deployed/available vs Trading's Account balance `[MVP]`
- [ ] 6.3 Working/in-flight order reconciliation vs Trading's open orders `[MVP]`
- [ ] 6.4 Mismatch handling: alert, reduce-only, state diff logging `[MVP]`
- [ ] 6.5 In-flight-aware comparison: downgrade mismatches explainable by pending orders to WARNING `[MVP]`

#### Phase 7 — Strategy Runner (~3.5d: 2 MMVP, 1.5 MVP)

- [ ] 7.1 Manifest loading and validation (capital_pool, capital_pct sum, .py files, predictor_fn IDs) `[MMVP]`
- [ ] 7.2 Strategy class instantiation, on_load/on_save lifecycle `[MMVP]`
- [ ] 7.3 Event dispatch loop: on_signal, on_outcome, on_timer, on_startup, on_shutdown `[MMVP]`
- [ ] 7.4 Serialized callback execution (one at a time per strategy) `[MMVP]`
- [ ] 7.5 Timer registration (multiple timers per strategy with timer_id) `[MVP]`
- [ ] 7.6 Exception handling: ABORT + EXIT + halt strategy, other strategies continue `[MVP]`
- [ ] 7.7 Periodic state checkpoints (on_save at STATE_CHECKPOINT_INTERVAL_SECONDS) `[MVP]`
- [ ] 7.8 Strategy template + SDK types (BaseStrategy protocol, Context, Action, Signal, Outcome dataclasses, single-file template with all callbacks stubbed and typed) `[MVP]`

#### Phase 8 — Hot Reload (~2.5d: 2.5 MVP)

- [ ] 8.1 Manifest file watching and diff detection `[MVP]`
- [ ] 8.2 Tier 1 changes: capital_pool, capital_pct, params, risk_limits, timers `[MVP]`
- [ ] 8.3 Tier 2 changes: strategy add/remove/file change, predictor_fn add/remove `[MVP]`
- [ ] 8.4 Tier 3 validation: reject id changes `[MVP]`
- [ ] 8.5 Manifest validation on reload (same checks as startup) `[MVP]`

#### Phase 9 — Startup & Shutdown (~1.5d: 1.5 MMVP)

- [ ] 9.1 Startup sequence: load state → register → reconcile → load manifest → instantiate → on_load → predictor_fns → timers → set mode → on_startup `[MMVP]`
- [ ] 9.2 Shutdown sequence: stop signals → on_shutdown → submit through Validator (bypass health/mode) → wait terminal → on_save → persist → deregister `[MMVP]`
- [ ] 9.3 Crash-only design: shared startup/recovery code path `[MMVP]`

#### Phase 10 — Alert Routing (~1.5d: 1 MVP, 0.5 Later)

- [ ] 10.1 Per-severity routing: warning (log/dashboard), error (+ webhook), critical (+ SMS) `[MVP]`
- [ ] 10.2 Webhook integration (Slack/Discord/Telegram via URL) `[MVP]`
- [ ] 10.3 SMS integration for critical alerts `[Later]`

#### Phase 11 — Integration & Testing (~4.5d: 0.5 MMVP, 3 MVP, 1 Later)

- [ ] 11.1 Unit tests: Capital Controller (check-and-reserve, lifecycle, drawdown, per-strategy budgets) `[MMVP]`
- [ ] 11.2 Unit tests: Validator (all 6 stages, breach policy, per-strategy risk limits) `[MVP]`
- [ ] 11.3 Unit tests: Strategy Runner (event dispatch, exception handling, serialized callbacks) `[MVP]`
- [ ] 11.4 Integration: Praxis Connector E2E with Trading sub-system (mock) `[MVP]`
- [ ] 11.5 Integration: full startup → signal → validate → execute → outcome → state update cycle `[MVP]`
- [ ] 11.6 Integration: hot reload (all 3 tiers) `[MVP]`
- [ ] 11.7 Integration: misfire detection scenarios `[MVP]`
- [ ] 11.8 Integration: multi-strategy capital contention under load `[Later]`
- [ ] 11.9 Chaos tests: crash recovery, mid-trade restart `[Later]`

#### Totals

| Tier | Checkboxes | Days |
|------|-----------|------|
| MMVP | 28 | ~14d |
| MVP | 32 | ~16d |
| Later | 5 | ~2.5d |
| **Total** | **65** | **~32.5d** |

---

### Resolved Items

The following items from the original review have been resolved and incorporated:

- **Deadline Escalation**: Not a Manager concern. Praxis owns execution escalation (deadline management, order type switching based on fill progress). Strategy provides `deadline` + `execution_mode` + `order_type` — Praxis handles the rest. If a command expires, strategy receives EXPIRED and decides whether to resubmit
- **Predictor_fn Contract**: Incorporated into Signal Interpreter data surface. Signal schema (binary, future multi-class) defined in Named Concepts. Core layer is signal-schema-agnostic — gates on actions, not signals
- **LIMIT_MAKER as Venue-Side Safety Net**: Generalized to `maker_preference` parameter on the action surface (MAKER_ONLY / MAKER_PREFERRED / NO_PREFERENCE). MAKER_ONLY maps to post_only/LIMIT_MAKER on venue. Venue-agnostic, not Binance-specific
- **Capital — Compounding**: Position sizing method is strategy-authored logic. Strategy decides `size` on each action, Core layer validates against capital ceilings. `auto_compound` (per-strategy boolean in manifest) adjusts strategy_budget to include realized P&L. Advanced methods (Kelly, volatility-adjusted) remain strategy decisions

### Remaining

None. All items resolved and incorporated into the RFC.

blahmonkey:
## Summary

- Add frozen `StrategySpec` dataclass (strategy_id, file, predictor_fn_ids, capital_pct)
- Add frozen `Manifest` dataclass (capital_pool, strategies tuple)
- Add `load_manifest()` to parse YAML manifests with `allocated_capital` ceiling validation
- Add `_validate_strategy_files()` to verify .py files exist and have valid Python syntax
- Add `pyyaml>=6.0` runtime dependency

## Test plan

- [x] 21 tests for StrategySpec validation (empty fields, invalid types, bounds)
- [x] 15 tests for Manifest invariants (duplicate IDs, capital_pct sum)
- [x] 16 tests for load_manifest (YAML parsing, file validation, syntax errors)
- [x] Ruff lint: pass
- [x] Mypy strict: pass
- [x] 708 tests passing

Closes #16 (RFC-3001 Phase 7.1)

### DIFF
--- a/nexus/infrastructure/manifest.py
+++ b/nexus/infrastructure/manifest.py
@@ -1,0 +1,114 @@
+'''Manifest loading and validation for strategy configuration.
+
+Provides frozen dataclasses for manifest structure and a loader
+that parses YAML and validates all constraints including file
+existence and importability.
+'''
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from decimal import Decimal
+
+__all__ = ['Manifest', 'StrategySpec']
+
+_ZERO = Decimal('0')
+_ONE_HUNDRED = Decimal('100')
+
+
+@dataclass(frozen=True)
+class StrategySpec:
+    '''Immutable specification for a single strategy.
+
+    Args:
+        strategy_id: Unique identifier for this strategy.
+        file: Path to the Python strategy implementation.
+        predictor_fn_ids: Predictor IDs this strategy depends on (from Limen).
+        capital_pct: Capital allocation percentage for this strategy.
+    '''
+
+    strategy_id: str
+    file: str
+    predictor_fn_ids: tuple[str, ...]
+    capital_pct: Decimal
+
+    def __post_init__(self) -> None:
+        '''Validate strategy specification invariants.'''
+
+        if not isinstance(self.strategy_id, str) or not self.strategy_id.strip():
+            msg = 'StrategySpec.strategy_id must be a non-empty string'
+            raise ValueError(msg)
+
+        if not isinstance(self.file, str) or not self.file.strip():
+            msg = 'StrategySpec.file must be a non-empty string'
+            raise ValueError(msg)
+
+        if not isinstance(self.predictor_fn_ids, tuple) or not self.predictor_fn_ids:
+            msg = 'StrategySpec.predictor_fn_ids must be a non-empty tuple'
+            raise ValueError(msg)
+
+        for predictor_id in self.predictor_fn_ids:
+            if not isinstance(predictor_id, str) or not predictor_id.strip():
+                msg = 'StrategySpec.predictor_fn_ids must contain non-empty strings'
+                raise ValueError(msg)
+
+        if (
+            not isinstance(self.capital_pct, Decimal)
+            or not self.capital_pct.is_finite()
+        ):
+            msg = 'StrategySpec.capital_pct must be a finite Decimal'
+            raise ValueError(msg)
+
+        if self.capital_pct <= _ZERO or self.capital_pct > _ONE_HUNDRED:
+            msg = 'StrategySpec.capital_pct must be in (0, 100]'
+            raise ValueError(msg)
+
+
+@dataclass(frozen=True)
+class Manifest:
+    '''Immutable manifest for a Manager instance.
+
+    Args:
+        capital_pool: Total capital in quote asset for this instance.
+        strategies: Strategy specifications for this instance.
+    '''
+
+    capital_pool: Decimal
+    strategies: tuple[StrategySpec, ...]
+
+    def __post_init__(self) -> None:
+        '''Validate manifest invariants.'''
+
+        if (
+            not isinstance(self.capital_pool, Decimal)
+            or not self.capital_pool.is_finite()
+        ):
+            msg = 'Manifest.capital_pool must be a finite Decimal'
+            raise ValueError(msg)
+
+        if self.capital_pool <= _ZERO:
+            msg = 'Manifest.capital_pool must be positive'
+            raise ValueError(msg)
+
+        if not isinstance(self.strategies, tuple) or not self.strategies:
+            msg = 'Manifest.strategies must be a non-empty tuple'
+            raise ValueError(msg)
+
+        seen_ids: set[str] = set()
+        total_pct = _ZERO
+
+        for spec in self.strategies:
+            if not isinstance(spec, StrategySpec):
+                msg = 'Manifest.strategies must contain StrategySpec instances'
+                raise ValueError(msg)
+
+            if spec.strategy_id in seen_ids:
+                msg = f'Manifest.strategies contains duplicate strategy_id: {spec.strategy_id!r}'
+                raise ValueError(msg)
+
+            seen_ids.add(spec.strategy_id)
+            total_pct += spec.capital_pct
+
+        if total_pct > _ONE_HUNDRED:
+            msg = f'Manifest.strategies capital_pct sum {total_pct} exceeds 100'
+            raise ValueError(msg)
--- a/nexus/infrastructure/manifest.py
+++ b/nexus/infrastructure/manifest.py
@@ -7,10 +7,15 @@ existence and importability.
 
 from __future__ import annotations
 
+import importlib.util
 from dataclasses import dataclass
 from decimal import Decimal
+from pathlib import Path
+from typing import Any
 
-__all__ = ['Manifest', 'StrategySpec']
+import yaml
+
+__all__ = ['Manifest', 'StrategySpec', 'load_manifest']
 
 _ZERO = Decimal('0')
 _ONE_HUNDRED = Decimal('100')
@@ -112,3 +117,129 @@ class Manifest:
         if total_pct > _ONE_HUNDRED:
             msg = f'Manifest.strategies capital_pct sum {total_pct} exceeds 100'
             raise ValueError(msg)
+
+
+def load_manifest(path: Path, allocated_capital: Decimal) -> Manifest:
+    '''Load and validate a manifest from a YAML file.
+
+    Args:
+        path: Path to the YAML manifest file.
+        allocated_capital: Hard ceiling for capital_pool validation.
+
+    Returns:
+        Validated Manifest instance.
+
+    Raises:
+        ValueError: If manifest is invalid or capital_pool exceeds allocated_capital.
+        FileNotFoundError: If manifest file does not exist.
+    '''
+
+    if not path.exists():
+        msg = f'Manifest file not found: {path}'
+        raise FileNotFoundError(msg)
+
+    with path.open() as f:
+        data: dict[str, Any] = yaml.safe_load(f)
+
+    if not isinstance(data, dict):
+        msg = 'Manifest must be a YAML mapping'
+        raise ValueError(msg)
+
+    raw_capital_pool = data.get('capital_pool')
+    if raw_capital_pool is None:
+        msg = 'Manifest missing required field: capital_pool'
+        raise ValueError(msg)
+
+    capital_pool = Decimal(str(raw_capital_pool))
+
+    if capital_pool > allocated_capital:
+        msg = (
+            f'Manifest capital_pool {capital_pool} exceeds '
+            f'allocated_capital {allocated_capital}'
+        )
+        raise ValueError(msg)
+
+    raw_strategies = data.get('strategies')
+    if not isinstance(raw_strategies, list) or not raw_strategies:
+        msg = 'Manifest missing or empty strategies list'
+        raise ValueError(msg)
+
+    specs: list[StrategySpec] = []
+
+    for raw_spec in raw_strategies:
+        if not isinstance(raw_spec, dict):
+            msg = 'Each strategy must be a YAML mapping'
+            raise ValueError(msg)
+
+        strategy_id = raw_spec.get('id')
+        if strategy_id is None:
+            msg = 'Strategy missing required field: id'
+            raise ValueError(msg)
+
+        file = raw_spec.get('file')
+        if file is None:
+            msg = f'Strategy {strategy_id!r} missing required field: file'
+            raise ValueError(msg)
+
+        raw_predictor_fn_ids = raw_spec.get('predictor_fn_ids')
+        if not isinstance(raw_predictor_fn_ids, list) or not raw_predictor_fn_ids:
+            msg = f'Strategy {strategy_id!r} missing or empty predictor_fn_ids'
+            raise ValueError(msg)
+
+        predictor_fn_ids = tuple(raw_predictor_fn_ids)
+
+        raw_capital_pct = raw_spec.get('capital_pct')
+        if raw_capital_pct is None:
+            msg = f'Strategy {strategy_id!r} missing required field: capital_pct'
+            raise ValueError(msg)
+
+        capital_pct = Decimal(str(raw_capital_pct))
+
+        specs.append(
+            StrategySpec(
+                strategy_id=strategy_id,
+                file=file,
+                predictor_fn_ids=predictor_fn_ids,
+                capital_pct=capital_pct,
+            )
+        )
+
+    manifest = Manifest(capital_pool=capital_pool, strategies=tuple(specs))
+
+    _validate_strategy_files(manifest, path.parent)
+
+    return manifest
+
+
+def _validate_strategy_files(manifest: Manifest, base_path: Path) -> None:
+    '''Validate all strategy files exist and are valid Python.
+
+    Args:
+        manifest: Manifest with strategies to validate.
+        base_path: Base path for resolving relative file paths.
+
+    Raises:
+        ValueError: If any strategy file is missing or has syntax errors.
+    '''
+
+    for spec in manifest.strategies:
+        file_path = base_path / spec.file
+
+        if not file_path.exists():
+            msg = f'Strategy {spec.strategy_id!r} file not found: {file_path}'
+            raise ValueError(msg)
+
+        try:
+            spec_obj = importlib.util.spec_from_file_location(
+                spec.strategy_id,
+                file_path,
+            )
+            if spec_obj is None or spec_obj.loader is None:
+                msg = f'Strategy {spec.strategy_id!r} file cannot be loaded: {file_path}'
+                raise ValueError(msg)
+
+            module = importlib.util.module_from_spec(spec_obj)
+            spec_obj.loader.exec_module(module)
+        except SyntaxError as e:
+            msg = f'Strategy {spec.strategy_id!r} file has syntax error: {file_path}: {e}'
+            raise ValueError(msg) from e
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v0.18.0 on 30th of March, 2026
+
+- Add [`manifest.py`](nexus/infrastructure/manifest.py) with frozen `StrategySpec` dataclass (strategy_id, file, predictor_fn_ids, capital_pct) and frozen `Manifest` dataclass (capital_pool, strategies)
+- Add `load_manifest()` to parse YAML manifest files with `allocated_capital` ceiling validation
+- Add `_validate_strategy_files()` to verify strategy .py files exist and have valid Python syntax via `importlib`
+- Add `pyyaml>=6.0` runtime dependency for YAML manifest parsing
+- Add mypy override for yaml module missing type stubs
+- Add 52 tests covering StrategySpec validation, Manifest invariants, load_manifest parsing, and file validation (708 total)
+
 ## v0.17.0 on 28th of March, 2026
 
 - Add `StateStore` dependency to `OutcomeProcessor` for WAL persistence of risk events
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ authors = [
 ]
 license = { text = "MIT" }
 requires-python = ">=3.10"
-dependencies = ["structlog>=24.0", "orjson>=3.10", "msgpack>=1.0"]
+dependencies = ["structlog>=24.0", "orjson>=3.10", "msgpack>=1.0", "pyyaml>=6.0"]
 
 [project.optional-dependencies]
 dev = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,10 @@ no_implicit_optional = true
 module = ["msgpack", "msgpack.*"]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["yaml", "yaml.*"]
+ignore_missing_imports = true
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 python_files = ["test_*.py"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "vaquum-nexus"
-version = "0.17.0"
+version = "0.18.0"
 description = "Layer for intelligence and decision-making between signal generation and trade execution."
 readme = "README.md"
 authors = [
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import tempfile
 from decimal import Decimal
+from pathlib import Path
 
 import pytest
 
-from nexus.infrastructure.manifest import Manifest, StrategySpec
+from nexus.infrastructure.manifest import Manifest, StrategySpec, load_manifest
 
 
 class TestStrategySpec:
@@ -427,3 +429,319 @@ class TestManifest:
         )
 
         assert len(manifest.strategies) == 1
+
+
+def _write_yaml(path: Path, content: str) -> None:
+    '''Write YAML content to a file.'''
+
+    path.write_text(content)
+
+
+def _write_strategy_file(base: Path, rel_path: str, content: str = '') -> None:
+    '''Create a strategy .py file.'''
+
+    file_path = base / rel_path
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+    file_path.write_text(content or '# Strategy file\n')
+
+
+class TestLoadManifest:
+    '''Tests for load_manifest function.'''
+
+    def test_valid_manifest_loads(self) -> None:
+        '''Valid YAML manifest loads successfully.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp_path = Path(tmp)
+            path = tmp_path / 'manifest.yaml'
+
+            _write_strategy_file(tmp_path, 'strategies/momentum.py')
+            _write_strategy_file(tmp_path, 'strategies/mean_rev.py')
+
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: momentum
+    file: strategies/momentum.py
+    predictor_fn_ids:
+      - alpha_v1
+    capital_pct: 60
+  - id: mean_rev
+    file: strategies/mean_rev.py
+    predictor_fn_ids:
+      - sensor_vol
+    capital_pct: 40
+''',
+            )
+
+            manifest = load_manifest(path, Decimal('20000'))
+
+            assert manifest.capital_pool == Decimal('10000')
+            assert len(manifest.strategies) == 2
+            assert manifest.strategies[0].strategy_id == 'momentum'
+            assert manifest.strategies[1].strategy_id == 'mean_rev'
+
+    def test_file_not_found_raises(self) -> None:
+        '''Missing manifest file raises FileNotFoundError.'''
+
+        with pytest.raises(FileNotFoundError, match='Manifest file not found'):
+            load_manifest(Path('/nonexistent/manifest.yaml'), Decimal('10000'))
+
+    def test_capital_pool_exceeds_allocated_raises(self) -> None:
+        '''capital_pool > allocated_capital raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 15000
+strategies:
+  - id: test
+    file: test.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match='exceeds allocated_capital'):
+                load_manifest(path, Decimal('10000'))
+
+    def test_missing_capital_pool_raises(self) -> None:
+        '''Missing capital_pool raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+strategies:
+  - id: test
+    file: test.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing required field: capital_pool'):
+                load_manifest(path, Decimal('10000'))
+
+    def test_missing_strategies_raises(self) -> None:
+        '''Missing strategies raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(path, 'capital_pool: 10000\n')
+
+            with pytest.raises(ValueError, match='missing or empty strategies'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_empty_strategies_raises(self) -> None:
+        '''Empty strategies list raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies: []
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing or empty strategies'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_missing_id_raises(self) -> None:
+        '''Strategy missing id raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - file: test.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing required field: id'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_missing_file_raises(self) -> None:
+        '''Strategy missing file raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: test
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing required field: file'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_missing_predictor_fn_ids_raises(self) -> None:
+        '''Strategy missing predictor_fn_ids raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: test
+    file: test.py
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing or empty predictor_fn_ids'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_missing_capital_pct_raises(self) -> None:
+        '''Strategy missing capital_pct raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: test
+    file: test.py
+    predictor_fn_ids: [pred1]
+''',
+            )
+
+            with pytest.raises(ValueError, match='missing required field: capital_pct'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_malformed_yaml_raises(self) -> None:
+        '''Malformed YAML raises yaml.YAMLError.'''
+
+        import yaml
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(path, 'capital_pool: [unclosed')
+
+            with pytest.raises(yaml.YAMLError):
+                load_manifest(path, Decimal('20000'))
+
+    def test_non_mapping_yaml_raises(self) -> None:
+        '''Non-mapping YAML raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(path, '- item1\n- item2\n')
+
+            with pytest.raises(ValueError, match='must be a YAML mapping'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_duplicate_strategy_id_raises(self) -> None:
+        '''Duplicate strategy_id raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: same
+    file: test1.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 50
+  - id: same
+    file: test2.py
+    predictor_fn_ids: [pred2]
+    capital_pct: 50
+''',
+            )
+
+            with pytest.raises(ValueError, match='duplicate strategy_id'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_capital_pct_sum_over_100_raises(self) -> None:
+        '''capital_pct sum > 100 raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: a
+    file: a.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 60
+  - id: b
+    file: b.py
+    predictor_fn_ids: [pred2]
+    capital_pct: 50
+''',
+            )
+
+            with pytest.raises(ValueError, match='exceeds 100'):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_file_not_found_raises(self) -> None:
+        '''Missing strategy .py file raises ValueError with path.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp_path = Path(tmp)
+            path = tmp_path / 'manifest.yaml'
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: missing_file
+    file: nonexistent/strategy.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match=r"file not found.*nonexistent/strategy\.py"):
+                load_manifest(path, Decimal('20000'))
+
+    def test_strategy_file_syntax_error_raises(self) -> None:
+        '''Invalid Python syntax in strategy file raises ValueError.'''
+
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp_path = Path(tmp)
+            path = tmp_path / 'manifest.yaml'
+
+            strategy_file = tmp_path / 'bad_syntax.py'
+            strategy_file.write_text('def broken(\n')
+
+            _write_yaml(
+                path,
+                '''
+capital_pool: 10000
+strategies:
+  - id: bad_syntax
+    file: bad_syntax.py
+    predictor_fn_ids: [pred1]
+    capital_pct: 100
+''',
+            )
+
+            with pytest.raises(ValueError, match=r'syntax error.*bad_syntax\.py'):
+                load_manifest(path, Decimal('20000'))
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -1,0 +1,429 @@
+'''Tests for manifest loading and validation.'''
+
+from __future__ import annotations
+
+from decimal import Decimal
+
+import pytest
+
+from nexus.infrastructure.manifest import Manifest, StrategySpec
+
+
+class TestStrategySpec:
+    '''Tests for StrategySpec dataclass.'''
+
+    def test_valid_strategy_spec(self) -> None:
+        '''Valid StrategySpec creates successfully.'''
+
+        spec = StrategySpec(
+            strategy_id='momentum_v1',
+            file='strategies/momentum.py',
+            predictor_fn_ids=('cohort_alpha_v3', 'sensor_volatility'),
+            capital_pct=Decimal('25'),
+        )
+
+        assert spec.strategy_id == 'momentum_v1'
+        assert spec.file == 'strategies/momentum.py'
+        assert spec.predictor_fn_ids == ('cohort_alpha_v3', 'sensor_volatility')
+        assert spec.capital_pct == Decimal('25')
+
+    def test_strategy_spec_is_frozen(self) -> None:
+        '''StrategySpec is immutable.'''
+
+        spec = StrategySpec(
+            strategy_id='test',
+            file='test.py',
+            predictor_fn_ids=('pred1',),
+            capital_pct=Decimal('10'),
+        )
+
+        with pytest.raises(AttributeError):
+            spec.strategy_id = 'changed'  # type: ignore[misc]
+
+    def test_empty_strategy_id_raises(self) -> None:
+        '''Empty strategy_id raises ValueError.'''
+
+        with pytest.raises(ValueError, match='strategy_id must be a non-empty string'):
+            StrategySpec(
+                strategy_id='',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_whitespace_strategy_id_raises(self) -> None:
+        '''Whitespace-only strategy_id raises ValueError.'''
+
+        with pytest.raises(ValueError, match='strategy_id must be a non-empty string'):
+            StrategySpec(
+                strategy_id='   ',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_empty_file_raises(self) -> None:
+        '''Empty file raises ValueError.'''
+
+        with pytest.raises(ValueError, match='file must be a non-empty string'):
+            StrategySpec(
+                strategy_id='test',
+                file='',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_whitespace_file_raises(self) -> None:
+        '''Whitespace-only file raises ValueError.'''
+
+        with pytest.raises(ValueError, match='file must be a non-empty string'):
+            StrategySpec(
+                strategy_id='test',
+                file='  ',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_empty_predictor_fn_ids_raises(self) -> None:
+        '''Empty predictor_fn_ids raises ValueError.'''
+
+        with pytest.raises(ValueError, match='predictor_fn_ids must be a non-empty tuple'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=(),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_predictor_fn_ids_not_tuple_raises(self) -> None:
+        '''Non-tuple predictor_fn_ids raises ValueError.'''
+
+        with pytest.raises(ValueError, match='predictor_fn_ids must be a non-empty tuple'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=['pred1'],  # type: ignore[arg-type]
+                capital_pct=Decimal('10'),
+            )
+
+    def test_predictor_fn_ids_with_empty_string_raises(self) -> None:
+        '''predictor_fn_ids containing empty string raises ValueError.'''
+
+        with pytest.raises(ValueError, match='predictor_fn_ids must contain non-empty strings'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1', ''),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_predictor_fn_ids_with_whitespace_raises(self) -> None:
+        '''predictor_fn_ids containing whitespace-only string raises ValueError.'''
+
+        with pytest.raises(ValueError, match='predictor_fn_ids must contain non-empty strings'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1', '   '),
+                capital_pct=Decimal('10'),
+            )
+
+    def test_predictor_fn_ids_with_non_string_raises(self) -> None:
+        '''predictor_fn_ids containing non-string raises ValueError.'''
+
+        with pytest.raises(ValueError, match='predictor_fn_ids must contain non-empty strings'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1', 123),  # type: ignore[arg-type]
+                capital_pct=Decimal('10'),
+            )
+
+    def test_non_decimal_capital_pct_raises(self) -> None:
+        '''Non-Decimal capital_pct raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be a finite Decimal'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=25,  # type: ignore[arg-type]
+            )
+
+    def test_infinite_capital_pct_raises(self) -> None:
+        '''Infinite capital_pct raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be a finite Decimal'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('inf'),
+            )
+
+    def test_nan_capital_pct_raises(self) -> None:
+        '''NaN capital_pct raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be a finite Decimal'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('nan'),
+            )
+
+    def test_zero_capital_pct_raises(self) -> None:
+        '''Zero capital_pct raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be in'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('0'),
+            )
+
+    def test_negative_capital_pct_raises(self) -> None:
+        '''Negative capital_pct raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be in'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('-5'),
+            )
+
+    def test_capital_pct_over_100_raises(self) -> None:
+        '''capital_pct > 100 raises ValueError.'''
+
+        with pytest.raises(ValueError, match='capital_pct must be in'):
+            StrategySpec(
+                strategy_id='test',
+                file='test.py',
+                predictor_fn_ids=('pred1',),
+                capital_pct=Decimal('101'),
+            )
+
+    def test_capital_pct_exactly_100_allowed(self) -> None:
+        '''capital_pct of exactly 100 is allowed.'''
+
+        spec = StrategySpec(
+            strategy_id='test',
+            file='test.py',
+            predictor_fn_ids=('pred1',),
+            capital_pct=Decimal('100'),
+        )
+
+        assert spec.capital_pct == Decimal('100')
+
+    def test_capital_pct_fractional_allowed(self) -> None:
+        '''Fractional capital_pct is allowed.'''
+
+        spec = StrategySpec(
+            strategy_id='test',
+            file='test.py',
+            predictor_fn_ids=('pred1',),
+            capital_pct=Decimal('12.5'),
+        )
+
+        assert spec.capital_pct == Decimal('12.5')
+
+    def test_single_predictor_fn_allowed(self) -> None:
+        '''Single predictor_fn in tuple is allowed.'''
+
+        spec = StrategySpec(
+            strategy_id='test',
+            file='test.py',
+            predictor_fn_ids=('single_pred',),
+            capital_pct=Decimal('50'),
+        )
+
+        assert spec.predictor_fn_ids == ('single_pred',)
+
+    def test_multiple_predictor_fn_ids_allowed(self) -> None:
+        '''Multiple predictor_fn_ids are allowed.'''
+
+        spec = StrategySpec(
+            strategy_id='test',
+            file='test.py',
+            predictor_fn_ids=('pred1', 'pred2', 'pred3'),
+            capital_pct=Decimal('50'),
+        )
+
+        assert spec.predictor_fn_ids == ('pred1', 'pred2', 'pred3')
+
+
+def _make_spec(
+    strategy_id: str = 'test',
+    capital_pct: Decimal = Decimal('50'),
+) -> StrategySpec:
+    '''Create a valid StrategySpec for testing.'''
+
+    return StrategySpec(
+        strategy_id=strategy_id,
+        file='test.py',
+        predictor_fn_ids=('pred1',),
+        capital_pct=capital_pct,
+    )
+
+
+class TestManifest:
+    '''Tests for Manifest dataclass.'''
+
+    def test_valid_manifest(self) -> None:
+        '''Valid Manifest creates successfully.'''
+
+        spec1 = _make_spec('strategy_a', Decimal('60'))
+        spec2 = _make_spec('strategy_b', Decimal('40'))
+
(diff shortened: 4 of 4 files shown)


nexus/infrastructure/manifest.py
```diff
@@ -0,0 +1,245 @@
...
+            spec_obj = importlib.util.spec_from_file_location(
+                spec.strategy_id,
+                file_path,
+            )
+            if spec_obj is None or spec_obj.loader is None:
+                msg = f'Strategy {spec.strategy_id!r} file cannot be loaded: {file_path}'
+                raise ValueError(msg)
+
+            module = importlib.util.module_from_spec(spec_obj)
+            spec_obj.loader.exec_module(module)
```
zero-bang, line 242: `exec_module` fully executes the strategy module at load time — any module-level side effects (network connections, thread spawns, global state mutations) will run during manifest loading. The intent is syntax-only validation; use `compile()` instead:

```python
try:
    source = file_path.read_text()
    compile(source, str(file_path), 'exec')
except SyntaxError as e:
    msg = f'Strategy {spec.strategy_id!r} file has syntax error: {file_path}: {e}'
    raise ValueError(msg) from e
```

This also lets you drop the `importlib.util` import entirely.

nexus/infrastructure/manifest.py
```diff
@@ -0,0 +1,245 @@
...
+    if not isinstance(data, dict):
+        msg = 'Manifest must be a YAML mapping'
+        raise ValueError(msg)
+
+    raw_capital_pool = data.get('capital_pool')
+    if raw_capital_pool is None:
+        msg = 'Manifest missing required field: capital_pool'
+        raise ValueError(msg)
+
+    capital_pool = Decimal(str(raw_capital_pool))
```
zero-bang, line 153: `Decimal(str(raw_capital_pool))` raises `decimal.InvalidOperation` (not `ValueError`) when the YAML value is a non-numeric string like `capital_pool: "abc"`. The docstring only documents `ValueError` and `FileNotFoundError`. Wrap in a try/except:

```python
try:
    capital_pool = Decimal(str(raw_capital_pool))
except Exception as e:
    msg = f'Manifest capital_pool is not a valid number: {raw_capital_pool!r}'
    raise ValueError(msg) from e
```

Same issue on line 196 for `capital_pct`.

zero-bang (changes requested):
Two issues to fix before merging: (1) `_validate_strategy_files` fully executes each strategy module instead of just checking syntax — this is a correctness and safety bug. (2) Non-numeric YAML values for `capital_pool` / `capital_pct` raise `decimal.InvalidOperation` rather than the documented `ValueError`.

PR #1, opened 2026-03-16
ci: mirror Praxis CI pipeline

blahmonkey:
**NOTE:** The author must always **first review the PR themselves**, before requesting review from others, that means carefully having reviewed the full diff in GitHub.

## What does this PR change?

Mirrors Praxis CI pipeline into Nexus. Adds `pr_checks_mypy.yml` workflow for strict type checking. Updates `pr_checks_codeql.yml` and `pr_checks_tests.yml` to use `uv` for dependency installation, `setup-python@v5`, and `pytest -v` as the test runner (replacing `tests/run.py`). Adds `dev` extras to `pyproject.toml` with `pytest>=8.0` and `mypy>=1.10`, strict mypy config, and pytest config. Initializes `nexus/` package so all CI gates have a target. Updates project name to `vaquum-nexus` and repository URL.

## Checklist

- [x] I have reviewed full diff in "Files changed"
- [x] I left no unnecessary files in the changes
- [x] I ran `python tests/run.py` locally without errors (where applicable)
- [ ] I updated `/docs` (if behavior/API/config/user/etc changed)
- [x] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/dev-docs/blob/main/src/Writing-Docstrings.md) (for any changed public functions/classes)
- [ ] I updated CHANGELOG.md (unless only docs or other non-code aspect was changed)
- [ ] I updated pyproject.toml (unless only docs or other non-code aspect was changed)
- [x] I added and/or updated tests (if behavior changed or new code paths added)
- [x] I validated changes manually
- [x] I validated changes with LLM
- [x] I removed any extraneous examples/comments
- [ ] I linked issue to auto-close on merge (e.g., "Fixes #123") when applicable

PR #34, opened 2026-04-14
MMVP-X.1: Pipeline integration blockers

blahmonkey:

## What does this PR change?

Implements all MMVP-X.1 pipeline integration blockers — the work needed for end-to-end signal-to-trade flow between Limen, Nexus, and Praxis.

**Internal state integrity (X.1.1):**
- UTC-only timestamp enforcement across all domain types (TD-004)
- Whitespace-padded strategy_id rejection in StrategySpec (TD-017)
- Version-dispatched WAL codec deserialization (TD-001)
- Strategy event retention across checkpoint for rolling loss windows (TD-002)

**Lifecycle observability (X.1.6):**
- Replace bare bool returns with `LifecycleResult` + `FailureCategory` enum in CapitalController (TD-003)
- OutcomeProcessor invariant breaches now raise RuntimeError instead of silent False

**Signal flow (X.1.2):**
- New `SensorSpec` manifest schema replacing flat `permutation_ids`
- `_wire_sensors()` trains Limen Sensors via `Trainer(experiment_dir).train(permutation_ids)` (TD-009)
- `signal_producer.produce_signal()` for live feature preparation and predict-to-Signal translation
- `PredictLoop` timer-based signal generation with thread-safe scheduling
- `_stop_signals()` via PredictLoop.stop in ShutdownSequencer (TD-013)

**Command flow (X.1.3):**
- `PraxisOutbound` sync-to-async bridge via `asyncio.run_coroutine_threadsafe`
- Account registration/deregistration via PraxisOutbound (TD-005, TD-012)
- `_submit_actions()` wired with PraxisOutbound for shutdown EXIT/ABORT (TD-015)

**Outcome flow (X.1.4):**
- `PraxisInbound` queue-based TradeOutcome consumption
- `_wait_terminal()` polling PraxisInbound with configurable timeout (TD-016)

**Capital reconciliation (X.1.5):**
- `_reconcile_capital()` pulls Praxis positions, compares by trade_id, syncs position_notional (TD-006)
- Reconciled state persisted to WAL immediately

**Deferred to technical debt (TD-019 through TD-024):**
- TD-019: Cohort support (Limen Cohort not ready)
- TD-020: Experiment directory sandboxing per account
- TD-021: Market data provider (depends on Praxis TD-016 #3)
- TD-022: Sensor hot reload (depends on manifest watcher)
- TD-023: Action dataclass trade fields (blocks full command submission)
- TD-024: Praxis-only position import (no strategy_id mapping)

944 tests passing (up from 891). Adds `vaquum_limen>=1.52.0` as dependency. Bumps to v0.25.0.

## Checklist

- [x] I ran `python -m pytest` locally without errors (944 passing)
- [x] I updated `/docs` (TechnicalDebt.md updated with TD-019 through TD-024)
- [x] I updated CHANGELOG.md

### DIFF
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -86,6 +86,7 @@ class PraxisOutbound:
         try:
             command_id = future.result(timeout=self._timeout)
         except (TimeoutError, concurrent.futures.TimeoutError):
+            future.cancel()
             _log.error(
                 'submit_command timed out: command_id=%s',
                 command.command_id,
@@ -143,6 +144,7 @@ class PraxisOutbound:
         try:
             future.result(timeout=self._timeout)
         except (TimeoutError, concurrent.futures.TimeoutError):
+            future.cancel()
             _log.error('deregister timed out: account_id=%s', account_id)
             raise
 
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -74,7 +74,7 @@ class PraxisOutbound:
                 order_type=command.command_type,
                 execution_mode=None,
                 execution_params=None,
-                timeout=max(1, round(self._timeout)),
+                timeout=round(self._timeout),
                 reference_price=None,
                 maker_preference=None,
                 stp_mode=command.stp_mode,
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -9,7 +9,8 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures
 import logging
-from collections.abc import Awaitable, Callable
+from collections.abc import Callable, Coroutine
+from typing import Any
 
 from nexus.infrastructure.praxis_connector.trade_command import TradeCommand
 
@@ -34,11 +35,11 @@ class PraxisOutbound:
 
     def __init__(
         self,
-        submit_fn: Callable[..., Awaitable[str]],
+        submit_fn: Callable[..., Coroutine[Any, Any, str]],
         loop: asyncio.AbstractEventLoop,
         register_fn: Callable[[str], None] | None = None,
-        unregister_fn: Callable[[str], Awaitable[None]] | None = None,
-        pull_positions_fn: Callable[[str], dict[tuple[str, str], object]] | None = None,
+        unregister_fn: Callable[[str], Coroutine[Any, Any, None]] | None = None,
+        pull_positions_fn: Callable[[str], dict[tuple[str, str], Any]] | None = None,
         timeout: float = _DEFAULT_TIMEOUT,
     ) -> None:
         self._submit_fn = submit_fn
@@ -64,7 +65,7 @@ class PraxisOutbound:
 
         # NOTE: execution_mode and execution_params require Action fields (TD-023).
         # Placeholder values used until full Action → TradeCommand translation is built.
-        future = asyncio.run_coroutine_threadsafe(
+        future: concurrent.futures.Future[str] = asyncio.run_coroutine_threadsafe(
             self._submit_fn(
                 trade_id=command.trade_id or command.command_id,
                 account_id=command.account_id,
@@ -101,7 +102,7 @@ class PraxisOutbound:
             },
         )
 
-        return command_id
+        return str(command_id)
 
     def register_account(self, account_id: str) -> None:
         '''Register account with Praxis Trading.
@@ -136,7 +137,7 @@ class PraxisOutbound:
             msg = 'unregister_fn not configured'
             raise RuntimeError(msg)
 
-        future = asyncio.run_coroutine_threadsafe(
+        future: concurrent.futures.Future[None] = asyncio.run_coroutine_threadsafe(
             self._unregister_fn(account_id),
             self._loop,
         )
@@ -150,7 +151,7 @@ class PraxisOutbound:
 
         _log.info('account deregistered', extra={'account_id': account_id})
 
-    def pull_positions(self, account_id: str) -> dict[tuple[str, str], object]:
+    def pull_positions(self, account_id: str) -> dict[tuple[str, str], Any]:
         '''Pull positions snapshot from Praxis Trading.
 
         Args:
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -38,7 +38,7 @@ class PraxisOutbound:
         loop: asyncio.AbstractEventLoop,
         register_fn: Callable[[str], None] | None = None,
         unregister_fn: Callable[[str], Awaitable[None]] | None = None,
-        pull_positions_fn: Callable[[str], dict] | None = None,
+        pull_positions_fn: Callable[[str], dict[tuple[str, str], object]] | None = None,
         timeout: float = _DEFAULT_TIMEOUT,
     ) -> None:
         self._submit_fn = submit_fn
@@ -150,14 +150,14 @@ class PraxisOutbound:
 
         _log.info('account deregistered', extra={'account_id': account_id})
 
-    def pull_positions(self, account_id: str) -> dict:
+    def pull_positions(self, account_id: str) -> dict[tuple[str, str], object]:
         '''Pull positions snapshot from Praxis Trading.
 
         Args:
             account_id: Account identifier to query.
 
         Returns:
-            Detached positions snapshot from Praxis.
+            Positions keyed by (account_id, trade_id) tuples.
 
         Raises:
             RuntimeError: If pull_positions_fn is not configured.
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -59,7 +59,7 @@ class PraxisOutbound:
 
         Raises:
             TimeoutError: If Praxis does not respond within timeout.
-            RuntimeError: If the async call fails.
+            Exception: Propagates the original exception raised by submit_fn.
         '''
 
         # NOTE: execution_mode and execution_params require Action fields (TD-023).
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -7,6 +7,7 @@ from a sync Nexus thread.
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import logging
 from collections.abc import Awaitable, Callable
 
@@ -84,7 +85,7 @@ class PraxisOutbound:
 
         try:
             command_id = future.result(timeout=self._timeout)
-        except TimeoutError:
+        except (TimeoutError, concurrent.futures.TimeoutError):
             _log.error(
                 'submit_command timed out: command_id=%s',
                 command.command_id,
@@ -141,7 +142,7 @@ class PraxisOutbound:
 
         try:
             future.result(timeout=self._timeout)
-        except TimeoutError:
+        except (TimeoutError, concurrent.futures.TimeoutError):
             _log.error('deregister timed out: account_id=%s', account_id)
             raise
 
--- a/nexus/infrastructure/praxis_connector/praxis_outbound.py
+++ b/nexus/infrastructure/praxis_connector/praxis_outbound.py
@@ -61,6 +61,8 @@ class PraxisOutbound:
             RuntimeError: If the async call fails.
         '''
 
+        # NOTE: execution_mode and execution_params require Action fields (TD-023).
+        # Placeholder values used until full Action → TradeCommand translation is built.
         future = asyncio.run_coroutine_threadsafe(
             self._submit_fn(
                 trade_id=command.trade_id or command.command_id,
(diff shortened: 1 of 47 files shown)


nexus/startup/shutdown_sequencer.py
```diff
@@ -176,26 +201,62 @@ def _submit_actions(self) -> None:
...
+            )
+            return
+
+        for strategy_id, action in filtered:
+            _log.info(
+                'shutdown action pending submission',
+                strategy_id=strategy_id,
+                action_type=action.action_type.value,
+            )
+            self._submitted_command_ids.append(f'{strategy_id}:{action.action_type.value}')
```
zero-bang: **Bug: phantom command IDs cause `_wait_terminal` to always timeout.**

This line appends synthetic IDs like `"momentum_v1:EXIT"` to `_submitted_command_ids`, but `send_command()` is never called — no commands are actually submitted to Praxis. When `_wait_terminal` runs, it polls `PraxisInbound` looking for outcomes with `command_id in pending`, but no Praxis outcome will ever have `command_id == "momentum_v1:EXIT"`. The result: `_wait_terminal` spins for the full `shutdown_timeout` (default 120s) on every shutdown where `praxis_outbound` is configured and strategies return EXIT/ABORT actions.

Since commands can't be submitted yet (Action lacks trade fields per TD-023), this block should not populate `_submitted_command_ids` at all. Replace the loop body with a log message acknowledging that submission is blocked:

```python
_log.warning(
    'shutdown actions cannot be submitted yet (TD-023)',
    action_count=len(filtered),
)
```

Also: the test `test_submit_actions_filters_exit_abort` only exercises the path where `praxis_outbound is None`, so this bug is not caught. Consider adding a test with a mock `praxis_outbound` configured.

nexus/startup/sequencer.py
```diff
@@ -130,20 +170,100 @@ def _recover_state(self) -> None:
...
+
+        if old_notional != praxis_total_notional:
+            _log.warning(
+                'position_notional adjusted',
+                old=str(old_notional),
+                new=str(praxis_total_notional),
+            )
+            self._state.capital.position_notional = praxis_total_notional
+
+        if old_notional != praxis_total_notional:
```
zero-bang: **Bug: duplicate condition — checkpoint block should be inside the first `if`.**

This `if old_notional != praxis_total_notional:` on line 255 is identical to the one on line 247. Since `old_notional` and `praxis_total_notional` are not modified between the two checks, the second check is always true when reached. Merge them into a single block:

```python
if old_notional != praxis_total_notional:
    _log.warning(
        'position_notional adjusted',
        old=str(old_notional),
        new=str(praxis_total_notional),
    )
    self._state.capital.position_notional = praxis_total_notional

    try:
        self._state_store.checkpoint(self._state)
    except Exception as e:
        raise StartupError('reconcile_capital', f'checkpoint after reconciliation failed: {e}') from e
```

nexus/infrastructure/praxis_connector/praxis_outbound.py
```diff
@@ -0,0 +1,165 @@
...
+
+        future = asyncio.run_coroutine_threadsafe(
+            self._submit_fn(
+                trade_id=command.trade_id or command.command_id,
+                account_id=command.account_id,
+                symbol=command.symbol,
+                side=command.side,
+                qty=command.size,
+                order_type=command.command_type,
+                execution_mode=command.command_type,
```
zero-bang: **Bug: `command.command_type` is semantically wrong for `execution_mode`.**

`command.command_type` is a `TradeCommandType` (NEW_ORDER / AMEND_ORDER / CANCEL_ORDER). The `execution_mode` parameter describes trade execution strategy (SingleShot, Bracket, TWAP, etc. per TD-023). Passing `NEW_ORDER` as `execution_mode` will either be rejected by Praxis or cause incorrect execution behavior.

Since the execution mode is not yet available (Action lacks the field per TD-023), pass `None` here:

```python
order_type=command.command_type,
execution_mode=None,
```

zero-bang (changes requested):
Thorough work implementing the X.1 pipeline integration. The LifecycleResult refactor, UTC enforcement, WAL version dispatch, and Sensor wiring are all solid. Three issues need addressing before merge:

1. **`_submit_actions` populates phantom command IDs that cause `_wait_terminal` to always timeout** — no commands are actually sent to Praxis, but synthetic IDs are queued, so `_wait_terminal` will spin for the full `shutdown_timeout` (default 120s) waiting for outcomes that can never arrive.
2. **Duplicate condition in `_reconcile_capital`** — the same `if` check appears twice in sequence.
3. **`execution_mode=command.command_type` in PraxisOutbound** — maps a semantically wrong value to the Praxis parameter.

Forward-looking concern: once TD-023 is resolved and `_submit_actions` actually calls `send_command()`, the returned Praxis command IDs need to replace the synthetic IDs currently stored in `_submitted_command_ids`. The current structure will need to be reworked to store the real IDs — worth keeping in mind when designing TD-023.

blahmonkey:
## Review fixes pushed in 67ef47e

All findings addressed:

**zero-bang comments:**
1. **Phantom command IDs** (`shutdown_sequencer.py:217`) — Removed phantom ID registration. `_submit_actions` now logs that submission is pending TD-023 and does not append to `_submitted_command_ids`.
2. **Duplicate condition** (`sequencer.py:255`) — Merged into single `if` block with checkpoint nested inside.
3. **execution_mode wrong type** (`praxis_outbound.py:72`) — Set to `None` with NOTE comment. Also fixed `timeout` truncation: `max(1, round(self._timeout))`.

**Copilot comments:**
4. **Thread churn in PredictLoop** — Replaced unbounded `list[Timer]` with `dict[str, Timer]` (one per sensor). Added `_schedule_locked` requiring lock.
5. **_tick dispatches after stop** — Added second `_running` check under lock before `produce_signal`/`dispatch_signal`.
6. **_submit_actions never calls send_command** — Same as #1 above.
7. **int(timeout) truncation** — Same as #3 above.
8. **TD-009 stale** — Marked RESOLVED in TechnicalDebt.md.
9. **TD-013 stale** — Marked RESOLVED in TechnicalDebt.md.
10. **skipif reason mismatch** — Updated to match actual detection logic.
11. **Misleading test name** — Renamed to `test_invalid_experiment_dir_raises`.

blahmonkey:
## Review fixes round 2 — pushed in f15c0c1

**Copilot comments on latest push:**

1. **TimeoutError mismatch** (`praxis_outbound.py:92,146`) — Fixed. `concurrent.futures.TimeoutError` is a separate class from `builtins.TimeoutError` in Python 3.10 (aliased in 3.11+). Now catches both.
2. **_tick dispatches after stop** (`predict_loop.py:127`) — Already fixed in previous round (67ef47e). Second `_running` check under lock exists before `produce_signal`.
3. **Docstring mismatch** (`signal_producer.py:29`) — Fixed. Docstring now reflects that arbitrary Limen exceptions may propagate.
4. **int(True) bypasses bool validation** (`manifest.py:279`) — Fixed. Raw YAML values validated for type before coercion: bools and non-ints rejected explicitly.
5. **Relative experiment_dir depends on CWD** (`manifest.py:277`) — Fixed. Relative paths now resolved against `path.parent` (manifest directory).

blahmonkey:
## Review fixes round 3 — pushed in 434268a

1. **x_train[-1] selects column not row** (`signal_producer.py:47`) — Fixed. Changed to `x_train.tail(1).to_numpy()`.
2. **Stray temp dirs** (`test_strategy_loader.py:242`, `test_shutdown_sequencer.py:28`) — Fixed. Replaced `tempfile.mkdtemp()` with `tempfile.TemporaryDirectory()` which auto-cleans.
3. **structlog %s formatting** (`shutdown_sequencer.py:344`) — Fixed. Using keyword arg `account_id=self._account_id`.
4. **skipif too broad** (`test_signal_producer.py:38`) — Fixed. Module-level `pytestmark` replaced with `_needs_limen` marker on integration class only. `TestExtractValues` unit tests always run.
5. **Timer key collision** (`predict_loop.py:95`) — Fixed. Key is now `{strategy_id}:{sensor_id}`.
6. **send_command docstring** (`praxis_outbound.py:62`) — Fixed. Now says "Propagates the original exception raised by submit_fn".

blahmonkey:
## Review round 4 — no code changes needed

All 3 comments are on known deferred items already tracked:

1. **predict_loop.py:118** — Re-check before dispatch: A second `_running` check under lock already exists before `produce_signal` (added in 67ef47e). The window between that check and `dispatch_signal` is microseconds of in-process Python. Adding a third check would be redundant.

2. **shutdown_sequencer.py:177** — Submit not calling send_command: Intentionally stubbed. The warning log explicitly states TD-023 dependency. Tracked in TD-015 and TD-023. Addressed in review rounds 1 and 2.

3. **manifest.py:287** — Experiment path sandboxing: Tracked as TD-020. Requires per-account allowlists — production deployment concern, not MMVP scope.

blahmonkey:
## Review fixes round 5 — pushed in 6bf0c94

1. **Praxis Decimal coercion** (`sequencer.py:218`) — Fixed. `qty` and `avg_entry_price` now coerced via `Decimal(str(...))` before arithmetic.
2. **CHANGELOG overstates submit_actions** (`CHANGELOG.md:268`) — Fixed. Now reads "filtering; actual submission pending Action fields (TD-015, TD-023)".
3. **WAL truncate_keeping_events CRC failure** (`wal.py:177`) — No change. CRC failure during checkpoint is a real corruption signal — failing the checkpoint is correct. Silently swallowing it would mask data loss.

blahmonkey:
## Review fixes round 6 — pushed in 49dff43

1. **Cancel future on timeout** (`praxis_outbound.py:93,146`) — Fixed. `future.cancel()` called before re-raising on both `send_command` and `deregister_account` timeouts.
2. **WAL _v defaults to 0** (`wal_codec.py:78,417`) — Fixed. Missing `_v` now defaults to 1 (all pre-versioning payloads are v1 format). Test updated to verify backward compatibility.
3. **Event _v same issue** (`wal_codec.py:417`) — Fixed alongside #2.

blahmonkey:
## Review fixes round 7 — pushed in 288cdd2

1. **predict_loop _tick after stop** — No change. Same comment for the 4th time. Lock guard before produce_signal already exists (67ef47e).
2. **pull_positions typed as plain dict** (`sequencer.py:205`) — Fixed. Return type now `dict[tuple[str, str], object]`. Constructor type updated to match.
3. **Size comparison uses raw praxis_pos.qty** (`sequencer.py:237`) — Fixed. Now compares against the already-coerced `qty` Decimal.

blahmonkey:
## Review fixes round 8 — pushed in 907f190

1. **_wire_sensors duplicate on re-entry** (`sequencer.py:387`) — Fixed. `_wired_sensors.clear()` at start of method.
2. **sensor_id collision from folder name** (`sequencer.py:401`) — Fixed. sensor_id now uses `str(experiment_dir.resolve())` for full path uniqueness.
3. **Timeout mismatch** (`praxis_outbound.py:82`) — Fixed. Submit_fn timeout now uses `round(self._timeout)` matching the future wait.
4. **Thread churn in PredictLoop** (`predict_loop.py:96`) — No change. Valid scaling concern tracked for X.2 optimization. Current threading.Timer approach is correct for MMVP.

blahmonkey:
## Review fixes round 9 — pushed in cac94e1

1. **Full path in sensor_id** (`sequencer.py:393`) — No change. Full path is intentional for uniqueness across experiments with same folder name.
2. **Unguarded praxis_pos attributes** (`sequencer.py:220`) — Fixed. Wrapped qty/avg_entry_price access in try/except with per-position skip and warning.
3. **WAL truncate memory spike** (`wal.py:190`) — No change. Performance optimization tracked in TD-018/X.2.3.
4. **_wait_terminal drops non-terminal outcomes** (`shutdown_sequencer.py:245`) — No change. Non-terminal outcomes (ACK, PARTIAL) are expected during wait. Queue is per-account, no shared consumption.
5. **numpy scalar types dropped** (`signal_producer.py:78`) — Fixed. Added `np.generic` handling for `np.float64`, `np.int64` etc.
6. **Tuple unpacking unguarded** (`sequencer.py:268`) — Fixed. Key validated as 2-tuple before unpacking, wrapped in StartupError on failure.

blahmonkey:
## Review round 10 — pushed in e961c4b

1. **sensor_id machine-dependent** — No change. Internal ID, uniqueness is the priority. Already discussed in round 8.
2. **Timer tests use real sleep** — No change. Intentional — tests verify actual threading.Timer behavior. Event-based synchronization already used in `test_dispatches_signal`.
3. **sleep after stop** — No change. Same as #2.
4. **round(0.5) → 0 timeout** — Fixed. `max(1, round(self._timeout))` ensures minimum 1 second.
5. **Multi-element array float cast** — No change. Last element as float is intentional for latest prediction. Single-element int semantics preserved via `val.item()`.

blahmonkey:
## Review fixes round 11 — pushed in 256ae14

1. **Import order** (`test_shutdown_sequencer.py:16`) — Fixed. pytest moved to third-party group before local imports.
2. **WiredSensor docstring** (`sequencer.py:41`) — Fixed. Now says `{path_hash_12}:{permutation_id}`.

blahmonkey:
## Review round 12 — no code changes

1. **predict_loop _tick after stop** — No change. This is the 6th time this comment has appeared. A `_running` check under lock exists before `produce_signal()` (the slow I/O-bound call). The window between that check and `dispatch_signal()` is microseconds of in-process Python. Adding a redundant check before every method call is not warranted.

zero-bang (dismissed):

zero-bang (dismissed):

blahmonkey:
## Review fixes round 13 — pushed in 3aafffd

1. **Git URL vs versioned dependency** (`pyproject.toml:15`) — No code change. Git URL is intentional — Limen isn't versioned on PyPI. Fixed CHANGELOG to match (`git+https://...` instead of `>=1.52.0`).
2. **checkpoint vs append_mutation** (`sequencer.py:270`) — No change. Checkpoint is safe and correct. Optimization to use append_mutation is valid but not a bug.
3. **Signal docstring says timezone-aware** (`signal.py:59`) — Fixed. Now says "must be UTC".
4. **TradeOutcome docstring** (`trade_outcome.py:28`) — Fixed. Same.
5. **translate docstring** (`translate.py:55`) — Fixed. Same.

zero-bang (approved):

PR #15, opened 2026-03-24
feat: 3.4 price validation — staleness, spread, deviation, reference source

blahmonkey:
**NOTE:** The author must always **first review the PR themselves**, before requesting review from others, that means carefully having reviewed the full diff in GitHub.

## What does this PR change?

RFC 3.4 price validation implementation covering:
- Adds price-validation config fields to `InstanceConfig`: `book_staleness_max_seconds`, `max_spread_bps`, `price_deviation_max_bps`, `reference_price_source`
- Adds strict validation for config fields: positive staleness seconds, finite non-negative bps thresholds, allowed `reference_price_source`, and deviation-source requirement
- Adds `build_price_stage_limits_from_config(...)` to map `book_staleness_max_seconds` to `max_staleness_ms` and wire spread/deviation/source limits into Stage 3
- Adds reference-source identity to price contracts via `PriceCheckSnapshot.reference_price_source` and `PriceStageLimits.reference_price_source`
- Fixes deviation-stage behavior to deny deterministically when reference source is missing or mismatched
- Expands test coverage for config validation, source-aware deviation paths, consequence routing (stale => platform-critical, spread/deviation => strategy-warning)
- Bumps to v0.14.0

## Checklist

- [x] I have reviewed full diff in \"Files changed\"
- [x] I left no unnecessary files in the changes
- [x] I ran `python -m pytest` locally without errors (512 passing)
- [ ] I updated `/docs` (if behavior/API/config/user/etc changed)
- [x] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/dev-docs/blob/main/src/Writing-Docstrings.md) (for any changed public functions/classes)
- [x] I updated CHANGELOG.md
- [x] I updated pyproject.toml
- [x] I added and/or updated tests (if behavior changed or new code paths added)
- [x] I validated changes manually
- [x] I validated changes with LLM
- [x] I removed any extraneous examples/comments
- [ ] I linked issue to auto-close on merge (e.g., \"Fixes #123\") when applicable

PR #43, opened 2026-04-25
fix: drop Limen @v2.4.3 pin to match Praxis (always take latest Limen)

mikkokotila:
## Summary

- v0.30.0 pinned `vaquum_limen` to `@v2.4.3` to converge with Praxis. Praxis has since dropped the tag and now declares `vaquum_limen @ git+https://github.com/Vaquum/Limen` (untagged → master), so the two siblings now declare the same package with two different URL specifiers.
- Any consumer that pulls both Praxis and Nexus into one venv (e.g. `backtest_simulator`) trips pip's resolver with `ResolutionImpossible` at install time — pip treats distinct URL strings as distinct packages even when they point at the same repo at the same SHA.
- This drops the `@v2.4.3` ref so the URL matches Praxis's untagged form. pip unifies the two transitive specs, install resolves, and the operator policy "always take the latest Limen" is honoured in both Nexus and Praxis venvs.
- PATCH bump 0.30.0 → 0.30.1.

## Repro of the failure (from `backtest_simulator` PR #15 CI)

```
ERROR: Cannot install backtest_simulator and vaquum-praxis because these package versions have conflicting dependencies.

The conflict is caused by:
    vaquum-praxis 0.44.0 depends on vaquum-limen 2.4.3 (from git+https://github.com/Vaquum/Limen)
    vaquum-nexus 0.30.0 depends on vaquum-limen 2.4.3 (from git+https://github.com/Vaquum/Limen@v2.4.3)
```

Same Limen, two different URL strings → ResolutionImpossible.

## Test plan

- [ ] CI: pr_checks_ruff / pr_checks_mypy / pr_checks_tests / pr_checks_codeql green
- [ ] Post-merge: `pip install backtest_simulator[integration]` resolves on a clean Python 3.12 venv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.30.1 on 25th of April, 2026
+
+- Drop `@v2.4.3` ref from `vaquum_limen` dependency in `pyproject.toml`; the dependency now reads `vaquum_limen @ git+https://github.com/Vaquum/Limen`. v0.30.0 pinned Limen to `@v2.4.3` to converge with Praxis, but Praxis has since moved to the untagged form, so the two siblings declared the same package with two different URL specifiers — pip's resolver treats those as conflicting packages and fails install with `ResolutionImpossible` for any consumer (e.g. `backtest_simulator`) that pulls both Praxis and Nexus into one venv. Untagging here re-converges the URL strings so pip unifies them, and matches the operator policy "always take the latest Limen"
+
 ## v0.1.0 on 16th of March, 2026
 
 - Add CI pipeline mirroring Praxis: Ruff, Mypy strict, pytest, CodeQL workflows
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "vaquum-nexus"
-version = "0.30.0"
+version = "0.30.1"
 description = "Layer for intelligence and decision-making between signal generation and trade execution."
 readme = "README.md"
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.30.1 on 25th of April, 2026
```
zero-bang, line 3: This entry should be at the **bottom** of the file, after the v0.30.0 block (after line 332 in `main`). The existing CHANGELOG uses ascending chronological order — every entry from v0.1.0 through v0.30.0 is appended at the end. Placing v0.30.1 at the top breaks that convention.

Fix: move this entire block (lines 3–6 in the diff) to the end of the file, after the v0.30.0 section.

zero-bang (changes requested):
The fix itself is correct — dropping the `@v2.4.3` ref aligns the URL with Praxis's untagged form and resolves the `ResolutionImpossible` conflict. Version bump and CI are clean.

One issue: the CHANGELOG entry is placed at the top of the file instead of the bottom. This repo uses ascending chronological order (v0.1.0 through v0.30.0), so v0.30.1 should be appended after the v0.30.0 entry at the end of the file.

Forward-looking note: `nexus/startup/sequencer.py` lines 530 and 546 contain comments referencing `vaquum_limen 2.4.3` as the version these private-attribute accesses were validated against. Those comments are still useful as "last-verified" markers, but worth re-validating if a Limen upgrade lands that changes the private API surface.

mikkokotila:
Withdrawing — out of scope for me to PR against Nexus. Closing on operator instruction.

PR #51, opened 2026-05-01
feat(capital): max_allocation_per_trade_pct as CapitalController kwarg

mikkokotila:
## Summary

Make the per-trade allocation cap a constructor kwarg instead of an inlined module constant. Default preserves prior behaviour bit-for-bit.

## Problem

`MAX_ALLOCATION_PER_TRADE_PCT = Decimal('0.15')` was a module-level constant referenced inline at `check_and_reserve` (`capital_controller.py:296`). Operators running strategies whose intrinsic Kelly exceeds 15% of pool have no knob — every reservation gets denied at validation time, the order never reaches the venue, the decoder looks dead from the strategy's perspective. Surfacing the cap as a per-controller setting is the smallest change that makes this configurable without touching the validation flow itself.

## Change

- `CapitalController.__init__` gains `max_allocation_per_trade_pct: Decimal = MAX_ALLOCATION_PER_TRADE_PCT` keyword-only argument
- Constructor validates the kwarg is a positive `Decimal` (TypeError on wrong type, ValueError on `<= 0`)
- The check at `check_and_reserve` line 296 reads `self._max_allocation_per_trade_pct` instead of the module constant
- The module constant remains as the default value of the kwarg — no rename, no removal, no symbol change to the public surface

## Backward compatibility

Existing call sites:
- `praxis/launcher.py:1141`: `CapitalController(state.capital)` — unchanged, picks up the default
- bts `honesty/capital.py:131`: `CapitalController(state)` — unchanged
- All existing test fixtures using `CapitalController(cs)` — unchanged

## Tests

Seven new tests in `tests/test_capital_controller.py` under `TestMaxAllocationPerTradePctOverride`:

- \`test_default_matches_module_constant\` — pin backwards compatibility (`CapitalController(cs)` denies above-15% as before)
- \`test_higher_override_admits_previously_denied\` — `0.50` lets a 0.16 allocation through
- \`test_lower_override_denies_previously_admitted\` — `0.05` denies an at-default-limit allocation
- \`test_denial_reason_carries_override\` — failure message reflects the override value
- \`test_non_decimal_override_raises\` — `TypeError` on `float` input
- \`test_zero_override_raises\` — `ValueError` on `Decimal('0')`
- \`test_negative_override_raises\` — `ValueError` on `Decimal('-0.1')`

## Verification

- \`pytest tests/test_capital_controller.py -q\`: 107 passed
- \`pytest tests/ -q\`: 1261 passed, 8 skipped, 1 unrelated warning
[...]

PR #52, opened 2026-05-01
feat(strategy): produce_signal lookback kwarg + dtype-preserving extract

mikkokotila:
## Summary

Add a keyword-only `lookback: int = 1` argument to `produce_signal` so callers can feed N trailing prepared-feature rows into `sensor.predict` instead of just the last 1. Default 1 preserves the prior behaviour bit-for-bit. While editing the same module, fix a latent dtype-preservation bug in `_extract_values` that becomes reachable through the new path.

> **Stacked on top of #51.** Branched from `feat/capital-controller-allocation-cap-kwarg` so PR #51's commits are visible in the diff until that PR merges. After #51 lands on main, this PR's diff trims to just the changes here.

## Problem

`produce_signal` does:

```python
last_row = x_train.tail(1).to_numpy()
result = wired.sensor.predict({'x_test': last_row})
```

Stateful predictors (those with rolling adaptive thresholds, hysteresis state machines, or any state that evolves across rows) need a history of feature rows to function correctly. They work end-to-end under `uel.run` — which calls `predict` once with the full test set as `x_test` — but the runtime `PredictLoop` calls them with a single row every tick. State machines reset, rolling-window thresholds fall through to their no-history fallback, and the model fires far less often than the bundle's training-time backtest stats imply.

This isn't fringe. Hysteresis-based entry/exit, regime-aware adaptive thresholds, and online-learning gates are common production patterns for converting continuous probabilities to binary trade signals. Forcing them to a 1-row predict is the narrowing.

## Change

**`nexus/strategy/signal_producer.py`:**

1. Add `lookback: int = 1` keyword-only argument to `produce_signal`. Default 1 → `tail(1).to_numpy()` → identical to current behaviour. Callers opt in to history with `lookback=N`. The Signal still carries only the LAST row's `_preds` / `_probs` regardless of `lookback`. Caller-side validation: `TypeError` on non-int / `bool` (which subclasses `int`), `ValueError` on values < 1.

2. Fix dtype preservation in `_extract_values`. The multi-element branch previously cast the last value to `float` unconditionally:

   ```python
   else:
       values[key] = float(val[-1])
   ```

   Single-row callers always hit the `size == 1` branch which preserves int dtype. With the new lookback path producing N-element arrays for `_preds` (an int array), the bug becomes reachable and silently demotes the strategy-visible value from `int` to `float`. Now the last element's dtype is preserved regardless of array length, and zero-length arrays are skipped instead of indexing into them.

## Backward compatibility

- `nexus/strategy/predict_loop.py:133` calls `produce_signal(wired, market_data)` (no `lookback` kwarg) — uses the default 1, behaviour unchanged.
- No other callers of `produce_signal` in `nexus/` or `praxis/` (verified via `grep -rn 'produce_signal\b'`).
- Existing test `test_multi_element_array_takes_last` updated: assertion changes from `== 1.0` to `== 1` (the underlying integer equality holds either way; the change pins the new dtype preservation).
- Existing test `test_signal_values_are_scalars` (asserts `isinstance(val, (int, float))`) still passes — int still satisfies the disjunction.

## Tests

**`TestLookback` class** (7 tests, limen-marked; run in CI):

- `test_default_lookback_one_matches_baseline` — `produce_signal(wired, market_data)` and `produce_signal(wired, market_data, lookback=1)` produce identical Signal values
- `test_lookback_greater_than_one_invokes_predict_with_n_rows` — wraps `sensor.predict`, asserts `x_test.shape[0] == 10` when `lookback=10`
- `test_lookback_signal_carries_last_row` — fakes a predict that puts a marker only in the last row of the array; Signal carries that marker
- `test_lookback_zero_raises` — `ValueError`
- `test_lookback_negative_raises` — `ValueError`
- `test_lookback_non_int_raises` — `TypeError` on float
- `test_lookback_bool_raises` — `TypeError` on `bool` (rejected explicitly because `bool` is an `int` subclass)

**`TestExtractValues` additions** (3 new tests, run locally):

- `test_multi_element_int_array_preserves_int` — int array of length > 1 → `int` (not `float`)
- `test_multi_element_float_array_yields_float` — float array of length > 1 → `float`
- `test_empty_array_skipped` — zero-length array doesn't add the key

`test_multi_element_array_takes_last` updated: assertion `== 1.0` → `== 1` (now that int dtype is preserved).

## Verification

- `pytest tests/test_signal_producer.py -q`: 7 passed, 10 skipped (limen-marked tests skip locally without datasets adjacent; pass in CI).
- `pytest tests/ -q`: 1264 passed, 15 skipped, 1 unrelated warning. Delta from main + #51 baseline (1261 / 8): +3 passed (`TestExtractValues` additions), +7 skipped (`TestLookback` need limen datasets locally).
- `mypy nexus/strategy/signal_producer.py`: clean.
- `ruff check` on touched files: clean.

## Version

`pyproject.toml`: 0.35.0 → 0.36.0
`CHANGELOG.md`: `v0.36.0 on 1st of May, 2026` entry under the existing format

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nexus/strategy/signal_producer.py
+++ b/nexus/strategy/signal_producer.py
@@ -14,20 +14,34 @@ from nexus.strategy.signal import Signal
 __all__ = ['produce_signal']
 
 
-def produce_signal(wired: WiredSensor, market_data: pl.DataFrame) -> Signal:
+def produce_signal(
+    wired: WiredSensor,
+    market_data: pl.DataFrame,
+    *,
+    lookback: int = 1,
+) -> Signal:
     '''Run feature preparation and predict to produce a Signal.
 
     Args:
         wired: Trained Sensor with limen_manifest and round_params.
         market_data: Rolling DataFrame of market bars.
+        lookback: Trailing prepared rows fed to sensor.predict. Default 1.
 
     Returns:
         Signal with predict output as values.
 
     Raises:
-        ValueError: If market_data is empty or x_train is missing.
+        TypeError: If lookback is not int.
+        ValueError: If market_data is empty, x_train is missing, or lookback < 1.
         Exception: Arbitrary exceptions from Limen prepare_data or predict.
     '''
+
+    if not isinstance(lookback, int) or isinstance(lookback, bool):
+        msg = f'lookback must be int, got {type(lookback).__name__}'
+        raise TypeError(msg)
+    if lookback < 1:
+        msg = f'lookback must be >= 1, got {lookback}'
+        raise ValueError(msg)
 
     if market_data.is_empty():
         msg = f'market_data is empty for sensor {wired.sensor_id}'
@@ -44,9 +58,9 @@ def produce_signal(wired: WiredSensor, market_data: pl.DataFrame) -> Signal:
         msg = f'prepare_data returned no x_train for sensor {wired.sensor_id}'
         raise ValueError(msg)
 
-    last_row = x_train.tail(1).to_numpy()
+    tail_rows = x_train.tail(lookback).to_numpy()
 
-    result = wired.sensor.predict({'x_test': last_row})
+    result = wired.sensor.predict({'x_test': tail_rows})
 
     values = _extract_values(result)
 
@@ -61,7 +75,8 @@ def _extract_values(result: dict[str, Any]) -> dict[str, Any]:
     '''Convert predict output to Signal-compatible values dict.
 
     Extracts scalar values from numpy arrays. Skips private keys
-    that are not signal values.
+    that are not signal values. Multi-element arrays use the last
+    element with dtype preserved.
     '''
 
     values: dict[str, Any] = {}
@@ -71,11 +86,10 @@ def _extract_values(result: dict[str, Any]) -> dict[str, Any]:
             continue
 
         if isinstance(val, np.ndarray):
-            if val.size == 1:
-                scalar = val.item()
-                values[key] = int(scalar) if isinstance(scalar, (np.integer, int)) else float(scalar)
-            else:
-                values[key] = float(val[-1])
+            if val.size == 0:
+                continue
+            last = val.item() if val.size == 1 else val[-1]
+            values[key] = int(last) if isinstance(last, (np.integer, int)) else float(last)
         elif isinstance(val, np.generic):
             values[key] = int(val) if isinstance(val, np.integer) else float(val)
         elif isinstance(val, (int, float)):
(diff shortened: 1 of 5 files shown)


zero-bang (dismissed):
Clean, well-structured change. Reviewed the `lookback` parameter addition, the `_extract_values` dtype-preservation fix, validation logic, caller impact (only `predict_loop.py:133` — uses default, unchanged), and test coverage.

**Correctness**: `x_train.tail(lookback)` with default 1 is bit-for-bit identical to the prior `tail(1)`. The `_extract_values` fix correctly handles the size==1 (`val.item()` → Python scalar) vs size>1 (`val[-1]` → numpy scalar) paths — the `isinstance(last, (np.integer, int))` check covers both branches. Empty-array skip is a clean guard against an edge that the old code would have blown up on.

**Backward compat**: Keyword-only with default 1 means the sole production caller (`predict_loop.py:133`) is unaffected. `__all__` doesn't change. Signal shape doesn't change.

**Tests**: Seven lookback tests (limen-gated, will run in CI) cover the key paths — default-matches-explicit, N-row predict shape, last-row-into-signal, and all four validation arms. Three new `_extract_values` tests pin the dtype fix directly.

**Forward-looking note**: if a caller passes `lookback=N` where N exceeds `x_train` height, Polars `tail()` silently returns all available rows. This is reasonable tail semantics but worth keeping in mind — a stateful predictor expecting exactly N rows of history might behave differently with fewer. Not a blocker; just the next thing to watch in production.

nexus/strategy/signal_producer.py
```diff
@@ -61,7 +80,9 @@ def _extract_values(result: dict[str, Any]) -> dict[str, Any]:
     '''Convert predict output to Signal-compatible values dict.
 
     Extracts scalar values from numpy arrays. Skips private keys
-    that are not signal values.
+    that are not signal values. The last element of multi-row
+    predictions is the runtime tick's value; dtype is preserved
+    so int arrays (`_preds`) stay int regardless of array length.
```
mikkokotila: The docstrings are not for prose. See what what the pre-change is and consider what is the minimal update for this docstring as per the new change.

nexus/strategy/signal_producer.py
```diff
@@ -14,21 +14,40 @@
...
 
     Args:
         wired: Trained Sensor with limen_manifest and round_params.
         market_data: Rolling DataFrame of market bars.
+        lookback: Number of trailing prepared-feature rows to feed into
+            `sensor.predict`. Default 1 preserves the prior single-row
+            behaviour bit-for-bit. Stateful predictors (rolling adaptive
+            thresholds, hysteresis state machines) that need history
+            across rows pass `lookback=N`. The Signal carries the LAST
+            row's `_preds` / `_probs` regardless of `lookback`.
```
mikkokotila: The docstrings and especially the args section is not a place for propose. Follow the form that is already there. Make your case in the PR body and not in docstrings. Leave the mvp blurb only in docstrings.

zero-bang (approved):

PR #61, opened 2026-05-06
feat(strategy): PredictLoop.tick_once synchronous single-shot entry point

mikkokotila:
## Summary

- Add public `PredictLoop.tick_once(wired)` for schedule-driven callers (deterministic backtest replay) that own their own clock and call it exactly at each pre-computed kline boundary
- Mirrors the `tick_once()` shape already on `OutcomeLoop` and `HealthLoop`
- Pure addition: no existing line in `predict_loop.py` was modified — production Praxis callers see no behaviour change

## Why

Backstory: bts (`backtest_simulator`) replay is non-deterministic today because its launcher's main thread `freezer.tick(...)` and Nexus's `PredictLoop` `threading.Timer` poll loop race on the same frozen `datetime.now(UTC)`. Same `(decoder, day, bundle)` produces different P&L between runs because the Timer thread fires at frozen instants the main thread has already shifted past. The `produce_signal` capture sees identical model probabilities at *shifted* timestamps and sometimes drops boundaries entirely — a real "strategy tested ≠ strategy deployed" violation.

The structural fix is to drive the predict cycle from a deterministic single-thread schedule in bts. To do that without forking Nexus semantics into a parallel universe, Nexus needs a public single-shot entry that the schedule driver can call. This PR adds exactly that and nothing else.

## What changes

- `nexus/strategy/predict_loop.py`: append `tick_once(wired)`. Body duplicates the chain Timer-driven `_tick` runs (market_data → produce_signal → dispatch_signal → optional action_submit). Duplication over extraction is deliberate so `_tick` stays byte-identical.
- `tests/test_predict_loop_tick_once.py`: 12 tests including a parity guard (`test_chain_call_order_matches_timer_tick`) that pins the chain call order so future edits to `_tick` can't drift the duplicated body silently.
- `pyproject.toml`: 0.40.0 → 0.41.0 (semver minor for new public API).
- `CHANGELOG.md`: v0.41.0 entry.

Differences from `_tick`: no `start()` requirement, no follow-up Timer scheduling, exceptions propagate to the caller (no swallow), and a `RuntimeError` guard that the two paths cannot interleave.

## Test plan

- [x] `pytest tests/test_predict_loop_tick_once.py` — 12/12 pass
- [x] `pytest tests/test_predict_loop.py` — 9/9 still pass (pure-additivity check)
- [x] `ruff check nexus/strategy/predict_loop.py tests/test_predict_loop_tick_once.py` — clean
- [x] `mypy nexus/strategy/predict_loop.py` — clean
- [ ] CI passes on this PR

## Consumer

After merge + release, [backtest_simulator](https://github.com/Vaquum/backtest_simulator) bumps its `vaquum-nexus>=0.41.0` floor and replaces its `_advance_clock_until` tick-and-sleep loop with a schedule-driven driver that calls `predict_loop.tick_once(wired)` per sensor at each pre-computed boundary. That is its own slice with its own PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/test_predict_loop_tick_once.py
```diff
@@ -0,0 +1,353 @@
...
+    interval_seconds: int = 1,
+) -> WiredSensor:
+    return WiredSensor(
+        sensor_id=sensor_id,
+        sensor=_mock_sensor(),
+        limen_manifest=_mock_limen_manifest(),
+        round_params={'random_weights': 0.5},
+        strategy_id=strategy_id,
+        interval_seconds=interval_seconds,
+    )
```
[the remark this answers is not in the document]
mikkokotila: We are not doing a test refactor in this PR.

tests/test_predict_loop_tick_once.py
```diff
@@ -0,0 +1,412 @@
...
+            market_data_provider=_mock_market_data_provider,
+            context_provider=_mock_context_provider,
+        )
+
+        loop.tick_once(wired)
+        loop.tick_once(wired)
+
+        assert runner.dispatch_signal.call_count == 2
+        assert loop._active_timers == {}
+
```
bit-mis, line 318: The docstring (and the PR description) promise this test catches drift when `_tick` is edited — but the body only exercises `tick_once`. `_tick` is never invoked and the two paths' call orders are never compared. If `_tick` is later edited to swap, drop, or reorder steps in the chain, this test will still pass. The parity rationale that justifies the deliberate duplication therefore isn't delivered. Drive `_tick` directly (e.g., set `_running=True` and patch `_schedule_locked` to a no-op so it doesn't recurse) using the same `tracking_market_data_provider`/`tracking_context_provider`, then assert the recorded chain order from `_tick` matches the one recorded from `tick_once`.

nexus/strategy/predict_loop.py
```diff
@@ -92,6 +92,49 @@ def stop(self) -> None:
...
+
+        Args:
+            wired: The wired sensor to fire one predict tick on.
+
+        Raises:
+            RuntimeError: When called while the Timer-driven loop is
+                running.
+        '''
+
+        if self._running:
```
bit-mis: `self._running` is read here without `self._lock`, while `start`, `stop`, and `_tick` all touch it under the lock. This is a TOCTOU race: a caller can pass this check reading `False`, and before the rest of `tick_once` runs `start()` acquires the lock, flips `_running` to `True`, and schedules timers — both paths then interleave, which the `Raises: RuntimeError ... while the Timer-driven loop is running` clause above states is impossible. Either acquire `self._lock` around the check (mirroring `_tick`'s pattern) or soften the docstring claim to reflect that the guard is best-effort.

bit-mis (changes requested):
Change requested in inline comments.

zero-bang (dismissed):
LGTM — clean addition. The `tick_once` body correctly mirrors `_tick`'s core chain (`_extract_kline_size` → `market_data_provider` → `produce_signal` → `context_provider` → `dispatch_signal` → `action_submit`), and the intentional differences (no scheduling, no exception swallowing, RuntimeError interleave guard) are well-motivated for the deterministic-backtest use case.

The `inspect.getsource`-based drift sentinel (`test_chain_markers_present_and_ordered_in_both_paths`) is a nice touch — it catches chain divergence in either direction at CI time, which is the right tradeoff for deliberate duplication over extraction.

Tests are thorough: 13 methods covering the happy path, all three exception-propagation paths, empty-data short-circuit, action_submit presence/absence, the RuntimeError guard, repeated independent calls, source-level parity, and runtime call ordering.

Minor: the CHANGELOG entry says "12 tests" but the file contains 13 test methods.

tests/test_predict_loop_tick_once.py
mikkokotila: Addressed in fc548a2 — added  which drives both  (with `_running` flipped and `_schedule_locked` patched via `unittest.mock.patch.object` to a no-op so it does not recurse) and `tick_once` with identical tracking providers, then asserts the recorded event lists match byte-for-byte. The existing static-source sentinel is kept as a complementary check.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- Add [`tests/test_predict_loop_tick_once.py`](tests/test_predict_loop_tick_once.py) — 12 tests covering: synchronous dispatch without `start()`, no follow-up Timer scheduling, `RuntimeError` guard against concurrent Timer-loop use, empty-market-data short-circuit, action_submit forwarding (called/not-called/exception-propagated branches), exception propagation from market_data_provider and dispatch_signal, repeated independent calls, and a parity guard that pins the chain call order so future edits to `_tick` cannot drift the duplicated `tick_once` body silently
+
```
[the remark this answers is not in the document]
mikkokotila: Addressed in 9e76409 — moved the v0.41.0 entry to its correct chronological position after v0.40.0.

nexus/strategy/predict_loop.py
```diff
@@ -92,6 +92,49 @@ def stop(self) -> None:
...
+            msg = 'tick_once must not be called while the Timer-driven loop is running'
+            raise RuntimeError(msg)
```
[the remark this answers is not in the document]
mikkokotila: Addressed in fc548a2 — the `_running` check is now wrapped in `with self._lock:` so the read is atomic with `start()`/`stop()` (no race between the guard and `start()` flipping the flag). Docstring updated to state honestly that the chain body still runs without the lock, so callers must not invoke `start()` mid-tick.

nexus/strategy/predict_loop.py
mikkokotila: Addressed in fc548a2 — wrapped the `_running` check in `with self._lock:` (mirroring `_tick`'s pattern). Updated the docstring to reflect that the guard is atomic at entry but the chain body runs without the lock, so concurrent `start()` is the caller's responsibility to avoid (rather than overstating it as a hard interleave barrier).

tests/test_predict_loop_tick_once.py
mikkokotila: Correction to my previous reply (which got partially eaten by shell quoting): the new test is named `test_chain_invoked_in_same_order_as_tick`. It drives both `_tick` (with `_running` flipped manually and `_schedule_locked` patched via `unittest.mock.patch.object` to a no-op so it does not reschedule a Timer) and `tick_once` against identical tracking providers, then asserts the recorded event lists match byte-for-byte. A future edit that changes the chain in only one of the two methods now fails CI here.

zero-bang (dismissed):

bit-mis (approved):

nexus/strategy/predict_loop.py
```diff
@@ -92,6 +92,54 @@ def stop(self) -> None:
...
+                so it is atomic with `start()` and `stop()`, but the
+                chain body runs without the lock; callers therefore
+                must not invoke `start()` while a `tick_once` chain
+                is in flight.
+        '''
+
+        with self._lock:
+            if self._running:
+                msg = 'tick_once must not be called while the Timer-driven loop is running'
+                raise RuntimeError(msg)
```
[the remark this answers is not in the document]
mikkokotila: Addressed in f6bf3a8 — closing the in-flight race in code would require modifying `_tick` to track in-flight count, which violates this PR's pure-additive constraint (the explicit goal: zero risk to existing Timer-driven callers including production Praxis). Documented the limitation honestly instead: the `Raises:` section now calls out both that the chain body runs without `_lock` (so callers must not invoke `start()` mid-tick) and that `stop()` does not wait for in-flight `_tick` (so callers must ensure any executing `_tick` has returned before invoking `tick_once`). Schedule-driven callers — the intended use case — call `tick_once` from a single thread that never touches the Timer-driven loop, so neither concern applies in practice; the docstring spells that out so future readers understand the design boundary.

zero-bang (approved):

mikkokotila:
CI Tests run on `f6bf3a8` failed on a **flaky pre-existing test** unrelated to this PR:

```
FAILED tests/test_outcome_processor.py::TestPositionsLockHonoredByWriter::test_concurrent_iteration_with_writer_does_not_raise
- OSError: [Errno 39] Directory not empty: 'wal'
```

Failure mode: the test spawns a writer thread that keeps appending WAL files while the foreground exits the `tempfile.TemporaryDirectory` context, and `rmtree` races with the writer's `wal/` directory creation. Last edited in `eec40d2` (pre-existing); none of the four files in this PR (`predict_loop.py`, `test_predict_loop_tick_once.py`, `pyproject.toml`, `CHANGELOG.md`) touch the outcome-processor test surface.

The 1353 other tests passed (15 skipped), including all 23 in the `predict_loop` suites. Locally on macOS this same test passes in 15s.

Triggered a `--failed` re-run.

PR #66, opened 2026-05-10
fix: support RuleBasedManifest in produce_signal predict path

blahmonkey:

## What does this PR change?

Fixes [`signal_producer.produce_signal`](nexus/strategy/signal_producer.py) to support `RuleBasedManifest` SFDs alongside `MLManifest`. The predict path was previously hard-coded against the `MLManifest.prepare_data` output shape — `data_dict.get('x_train')` for the live train frame and `wired.sensor.predict({'x_test': ...})` for the predict call. `RuleBasedManifest.prepare_data` returns a different shape: `{'train', 'val', 'test', '_alignment', 'strategy'}` with no `x_train` / `x_test` keys (per `RuleBasedManifest.prepare_data` → `_finalize_rule_based_data` → `split_data_to_rule_based_prep_output` in [Limen v3.0.6 manifest_core.py](https://github.com/Vaquum/Limen/blob/v3.0.6/limen/experiment/manifest_core.py)). So every rule-based sensor tick raised `ValueError: prepare_data returned no x_train for sensor <id>` on the empty-train guard, breaking the predict loop for the entire rule-based-SFD class.

Surfaced when a Praxis operator manifest YAML staged a `RuleBasedManifest`-backed stub-strategy bundle (`always_one`, `always_zero`, `alternating`, `coin_toss` — four trivial deployment-plumbing-validation stubs) alongside the existing `BtcLogRegEVSFD` (`MLManifest`) bundle. All four rule-based sensors crashlooped on the empty-train guard while the ML one ticked cleanly. Bumps `vaquum-nexus` 0.45.0 → 0.46.0.

**Fix.** Branch on `isinstance(wired.limen_manifest, RuleBasedManifest)` and pick the right keys for that path. ML-SFDs are unaffected — they keep using `x_train` / `x_test`. Imports `RuleBasedManifest` from `limen.experiment` (the canonical export, alongside `MLManifest`).

```python
is_rule_based = isinstance(wired.limen_manifest, RuleBasedManifest)
train_key = 'train' if is_rule_based else 'x_train'
test_key = 'test' if is_rule_based else 'x_test'

manifest_full = wired.limen_manifest.with_params_override(split_config=(1, 0, 0))
data_dict = manifest_full.prepare_data(market_data, wired.round_params)

train_frame = data_dict.get(train_key)
if train_frame is None or train_frame.is_empty():
    raise ValueError(f'prepare_data returned no {train_key!r} for sensor {wired.sensor_id}')

tail_frame = train_frame.tail(lookback)
result = wired.sensor.predict({test_key: tail_frame})
```

**Test.** Adds `_make_wired_rule_based_sensor` fixture (mirrors the existing `_make_wired_sensor` shape but drives `limen.sfd.foundational_sfd.rule_based` through `Trainer.train([1])` to produce a `WiredSensor` whose `limen_manifest` is a real `RuleBasedManifest`). Adds `test_predict_called_with_test_key_for_rule_based_manifest` under `TestLookback` — wraps `wired.sensor.predict` with a key-capturing shim, runs `produce_signal`, asserts (1) the captured `data` dict contains key `'test'` and (2) does NOT contain `'x_test'`. The negative assertion guards against a regression that would silently fall back to the ML key and crash inside the rule-based reference architecture's `predict(data)` which reads `data['test']`.

**Pre-PR pass.** `python -m pytest`: 1354 passed, 17 skipped (one new local skip — the rule-based fixture also uses `@_needs_limen` so it runs against Limen-installed-with-datasets in CI). `ruff check nexus tests`: clean. `mypy nexus/strategy/signal_producer.py`: clean.

## Checklist

- [x] I ran `python -m pytest` locally without errors (1354 passed, 17 skipped)
- [x] I updated CHANGELOG.md


zero-bang (approved):
Clean fix. Verified the `from limen.experiment import RuleBasedManifest` import against the pinned Limen v3.0.6 `__init__.py` — it's a canonical export alongside `MLManifest`. Confirmed `split_data_to_rule_based_prep_output` returns `{'train', 'val', 'test', '_alignment'}` (no `x_train`/`x_test`), so the key branching is correct.

Callers checked:

- `predict_loop.py` calls `produce_signal` transparently — no change needed. `_extract_kline_size` reads `data_source_config` which both manifest types inherit from `Manifest`.
- Mock-based tests in `test_predict_loop.py` / `test_predict_loop_tick_once.py` use `MagicMock()` for the manifest, which isn't an instance of `RuleBasedManifest`, so the existing ML path (`x_train`/`x_test`) is correctly taken.
- `WiredSensor.limen_manifest` is typed `Any`, so it holds whatever `trainer._manifest` yields — `RuleBasedManifest` or `MLManifest` depending on the SFD.

New test is solid — the negative assertion (`not has_x_test`) is the important half, catching a regression where the ML key silently survives into the rule-based path. Fixture mirrors the existing `_make_wired_sensor` pattern with a real `Trainer.train([1])` round-trip.

LGTM.

PR #67, opened 2026-05-13
fix: ExtraAdder in ProcessorFormatter + per-action bound_context (v0.47.0)

blahmonkey:

## What does this PR change?

Mirrors the Praxis-side fix in Vaquum/Praxis#105. Fixes the universal extras-drop bug in Nexus observability and adds per-action structlog correlation context to `submit_actions` so any downstream emit (validator stages, capital_controller, praxis_outbound, OutcomeProcessor) carries `strategy_id` / `action_type` / `trade_id` / `command_id` via `merge_contextvars` instead of having to thread them through `extra={...}` at every emit site. `configure_logging`'s `ProcessorFormatter.foreign_pre_chain` now includes `structlog.stdlib.ExtraAdder()` so stdlib `_log.info('msg', extra={...})` payloads merge into JSON output — pre-fix every Nexus emit site using `_log = logging.getLogger(__name__)` dropped its `extra` payload silently, including all 7 emit sites in `action_submit.py`, plus `outcome_processor.py`, `capital_controller.py`, `praxis_outbound.py`, and the validator stages. Adds `bound_context(**kwargs)` context manager wrapping `structlog.contextvars.bound_contextvars` so callers do not need to import structlog directly; `submit_actions`'s per-iteration body is now wrapped in `with bound_context(strategy_id, action_type, trade_id?, command_id?)` with reset on exit (including the exception path) so the caller's outer context is preserved across actions. Adds `test_stdlib_extras_appear_in_json` regression test, two `bound_context` lifecycle tests (binds/unbinds + reset-on-exception), and `TestPerActionBoundContext` (4 cases pinning per-action bind for ENTER/EXIT/ABORT). Bumps to v0.47.0.

## Checklist

- [x] I ran `python -m pytest` locally without errors (1360 passed, 17 skipped; one flaky `test_outcome_processor.py::TestPositionsLockHonoredByWriter::test_concurrent_iteration_with_writer_does_not_raise` re-ran in isolation and passed)
- [x] I updated CHANGELOG.md


zero-bang (dismissed):
Thorough read of the diff, the source files, all callers, and the structlog processor chain semantics. Ship it.

**ExtraAdder fix** — correct. stdlib `extra={...}` payloads were silently dropped because `ProcessorFormatter.foreign_pre_chain` lacked `ExtraAdder`. The fix puts it first in the chain so extras are in the event dict *before* `merge_contextvars` runs. Since `merge_contextvars` uses `setdefault`, explicit extras always take precedence over ambient contextvars — the right semantics.

**`bound_context`** — thin, correct wrapper consistent with the existing `bind_context`/`clear_context` API surface. Lifecycle (`__exit__` on `continue`, exception path) is well-tested.

**`submit_actions` wiring** — verified that:
- ABORT path correctly runs inside `bound_context` so `_submit_abort`'s downstream emits carry correlation fields.
- MODIFY actions bind `action.command_id` (the command being modified) in contextvars, while the "action submitted" log's `extra={'command_id': send_command_result}` (the new Praxis-assigned ID) wins via `setdefault` precedence. No semantic collision.
- ENTER/EXIT do not bind `command_id`/`trade_id` when absent. No spurious keys.
- `continue` inside `with bound_context(...)` properly triggers `__exit__`. Python guarantees this.

**Tests** — all 7 new tests are meaningful and correctly structured. Context isolation is clean (`clear_context()` bookends). The validator-capture pattern (side_effect reading `get_contextvars()`) is the right way to verify mid-iteration binding.

**Callers** — `PredictLoop`, `TimerLoop`, `OutcomeLoop` call via the `ActionSubmitter` callback. No signature change, no breakage. `ShutdownSequencer._submit_actions` is a separate method, unaffected.

No blocking issues.

zero-bang (dismissed):

zero-bang (approved):

PR #74, opened 2026-05-28
feat: predict-pool process executor and pre-launch wire-cache warmer

blahmonkey:

## What does this PR change?

Two related but separable runtime fixes, plus a pre-PR cleanup commit. Each is its own commit on the branch.

### `fix:` pre-launch sensor wire-cache warmer; inline-only wiring in launcher (v0.52.1)

The launcher owns Polars' global rayon thread pool (market-data cache). Creating a worker `ProcessPoolExecutor` in that process degrades the rayon pool, and a later large Polars merge **segfaults** the launcher (rayon worker stack overflow → SIGSEGV → restart-loop). Observed on a 5000-permutation deploy: `_polars_runtime.abi3.so` segfaults on a `polars-N` thread every ~37 min, fault address just below the thread stack pointer, regardless of worker count; 0.66.0 (serial, no pool) ran 8 h clean.

This PR moves parallel sensor reconstruction out of the launcher into a new pre-launch warmer (`python -m nexus.startup.warm_cache`). The warmer never imports Polars and exits before the launcher starts; the launcher then wires inline against the warm disk cache. CPU-bound reconstruction and the Polars-using trading runtime live in separate processes.

- **BREAKING**: `_wire_sensors` no longer runs a `ProcessPoolExecutor`; it reconstructs cache misses **inline only**. `NEXUS_WIRE_MAX_WORKERS` no longer affects the launcher — it now governs the standalone warmer. A deployment that previously set `NEXUS_WIRE_MAX_WORKERS > 1` to parallelize in-launcher wiring must move that to the pre-launch warmer step.
- **Add** `nexus/startup/warm_cache.py`: collects the union of `(experiment_dir, permutation_id)` tasks across one or more manifests (mirrors the launcher's `MANIFESTS_DIR` enumeration), skips already-cached entries, and reconstructs the rest across a `spawn` `ProcessPoolExecutor`. Pins `OMP_NUM_THREADS` / `OPENBLAS_NUM_THREADS` / `MKL_NUM_THREADS` / `NUMEXPR_NUM_THREADS` to `1` before spawning so `N` workers use `N` cores rather than `N × cores` (the BLAS oversubscription that made a 32-worker wire thrash all 48 cores at ~38 perms/min). Per-sensor `ReconstructionError` failures are logged and skipped.
- **Add** `tests/test_warm_cache.py`: the warmer dedups tasks within and across manifests, builds the pool with a `spawn` context, pins BLAS to one thread, and writes one cache entry per reconstructed sensor.

### `feat:` process-pool predict executor (spawn workers, shared market-data IPC) (v0.53.0)

The bounded `ThreadPoolExecutor` predict path is GIL-bound by Limen `MLManifest.prepare_data` (~4.4 s/predict, ~99.9% Python/Polars on a ~224k-row frame). On the 5000-permutation 15-minute deploy, threads don't scale; a full predict cycle took ~3.8 h and missed the 15-minute SLA. This PR moves the per-tick predict work into spawn worker processes (Codex-supervised audit signed off; deploy config measured at **6.37 predicts/sec** with 16 workers vs the required 5.05/sec).

Strategy dispatch, context lookup, and action submission stay in the parent process and are serialized by the pool's result-handling thread, so capital / WAL / risk plumbing is untouched.

- **Add** `experiment_dir: Path` field to `WiredSensor` (and pass `task.resolved_dir` at construction) so a predict worker can rebuild the (unpicklable) Limen manifest from disk via `Trainer(experiment_dir)._manifest` rather than receive it across the pickle boundary. The fitted `sensor` (small) still rides the submit pickle; the manifest is cached per worker in a module-level `_WORKER_MANIFESTS` dict.
- **Add** per-cycle Arrow IPC sharing of the aggregated market-data frame: the parent computes one aggregation per `(kline_size, latest_bar)`, writes it to a temp Arrow IPC file under a refcounted lifecycle (`_ipc_current`, `_ipc_refs`, `_maybe_unlink_ipc`), and submits each due sensor with `(PredictTask, path)`. Workers read the file (cached per path, bounded to 4 frames per worker). 5000 sensors sharing one kline_size pay one parent aggregation per cycle, not one per sensor.
- **Add** new worker surface in `nexus/strategy/predict_loop.py`: frozen `PredictTask` dataclass, `_predict_worker_init`, `_predict_in_process(task, market_data_path)` (rebuild manifest, read IPC, run `produce_signal`, return Signal), and `_market_data_signature` for the IPC cache key.
- **Add** env knobs `NEXUS_PREDICT_MAX_WORKERS` (default `16`) and `NEXUS_PREDICT_POLARS_MAX_THREADS` (default `1`); the latter is set via `os.environ.setdefault('POLARS_MAX_THREADS', ...)` before pool creation so spawned worker interpreters inherit it. Bounding Polars to one thread per worker stops BLAS-style oversubscription without changing parent-side Polars use.
- **Fix** circular-import blocker: `shutdown_sequencer.py`'s `from nexus.strategy.predict_loop import PredictLoop` moved under `TYPE_CHECKING`. The import was annotation-only (`from __future__ import annotations` already in place), but the existing chain `predict_loop → nexus.startup → shutdown_sequencer → predict_loop` prevented a spawn worker from importing `predict_loop` as a fresh entry point (the worker must, to unpickle `_predict_in_process`). The old in-process `ThreadPoolExecutor` never re-imported the module from a clean interpreter and so never hit the cycle.
- **Rework** `tests/test_predict_loop.py` to drive the parent scheduler via an in-process `_SyncExecutor` test double plus a stubbed `_predict_in_process` (autouse fixture scoped to `TestPredictLoop`). `MagicMock` sensors and manifests cannot cross a spawn pickle boundary or rebuild a manifest from a non-existent experiment dir, so testing parent-side behavior (dispatch chain, action_submit forwarding, empty-data skip, predict-error resilience, multi-sensor) needs the executor seam. New `TestPredictInProcess` class covers `_predict_in_process` directly by patching `Trainer` and writing a real IPC frame. `tick_once` is unchanged (still in-process synchronous), so its tests still drive `produce_signal` end-to-end with mocks.
- **Drop** two `_tick_worker`/`tick_once` drift-sentinel tests: the invariant they pinned (both paths run an identical step chain) no longer holds; the predict chain is now split between worker (`_predict_in_process → produce_signal`) and parent (`_handle_predict_result → _log_signal → dispatch_signal → _action_submit`).

### `fix:` pr-prep review pass (commit `8edc904`)

Greybeard pre-PR review caught five items, all addressed in commit 3:

- `predict_loop.py:1-17` module docstring rewritten to describe the spawn-pool / IPC design (the previous text still claimed a Timer-based, per-sensor-timer loop).
- `predict_loop.py:_predict_in_process` NOTE comment added beside the `Trainer(...)._manifest` private-attribute read, matching the established pattern at `sequencer.py:863-864`.
- `predict_loop.py:_maybe_unlink_ipc` `path.unlink(missing_ok=True)` wrapped in `try/except OSError` + logged exception. The function is called from the done-callback `finally`; an uncaught non-FNF unlink failure would propagate into the executor's result-handler thread where only the stdlib default handler would see it.
- `CHANGELOG.md` v0.52.1 and v0.53.0 sub-sections reordered so `Add` precedes `Fix` (matches the spec ordering `BREAKING → NOTE → Add → Fix → Update` and v0.52.0's pattern).
- `docs/TechnicalDebt.md` `TD-089` added recording the `_SyncExecutor` test-double limitation (fires `add_done_callback` inline rather than from a separate result-handler thread, so cross-thread races aren't exercised by tests).

## Verification

- `ruff check` — clean across changed modules
- `mypy nexus/strategy/predict_loop.py nexus/startup/shutdown_sequencer.py nexus/startup/sequencer.py` — clean
- `pytest tests/` — **1400 passed, 17 skipped** (the lone warning is the pre-existing WAL-teardown thread race in `test_outcome_processor.py`, unrelated)
- Real spawn-pool smoke test (`ProcessPoolExecutor(mp_context=spawn, initializer=_predict_worker_init)`) confirms workers import `predict_loop` cleanly and run — i.e. the circular-import fix holds in a fresh interpreter
- Production measurement (Codex-supervised audit, prod artifacts): 16-worker spawn pool clears 6.37 predicts/sec vs the 5.05/sec required for 4548 sensors on a 15-minute cadence

## Scope

`nexus/startup/sequencer.py`, `nexus/startup/shutdown_sequencer.py`, `nexus/startup/warm_cache.py` (new), `nexus/strategy/predict_loop.py`, `tests/test_predict_loop.py`, `tests/test_predict_loop_tick_once.py`, `tests/test_signal_producer.py`, `tests/test_warm_cache.py` (new), `CHANGELOG.md` (v0.52.1 + v0.53.0), `pyproject.toml` (`0.52.0 → 0.53.0`), `docs/TechnicalDebt.md` (TD-089).

## Checklist

- [x] I ran `python -m pytest` locally without errors (1391 passing, 17 skipped)
- [x] I updated `/docs` (if behavior/API/config/user/etc changed)
- [x] I updated CHANGELOG.md
- [x] I linked issue to auto-close on merge (e.g., "Fixes #123") when applicable


nexus/strategy/predict_loop.py
```diff
@@ -308,45 +534,202 @@ def tick_once(self, wired: WiredSensor) -> None:
...
-                    return
+            task = self._tasks[key]
+            self._ipc_refs[path] = self._ipc_refs.get(path, 0) + 1
+
+        wired = self._by_key[key]
+
+        def _on_done(future: Future[Signal]) -> None:
+            self._handle_predict_result(wired, path, future)
+
+        executor.submit(_predict_in_process, task, str(path)).add_done_callback(_on_done)
```
zero-bang: `executor.submit()` can raise `BrokenProcessPool` (a `RuntimeError` subclass) if any worker crashes (OOM, signal, segfault in a dependency). When that happens:

1. The IPC ref incremented at line 629 is never decremented (the done-callback never fires).
2. The `key` stays in `self._in_flight` permanently — subsequent polls skip it.
3. The exception propagates through `_submit_one` → `_submit_due` → `_scheduler_loop` (line 561, no `try/except`), killing the daemon scheduler thread silently. **All predictions stop with no log message.**

This is the only unguarded `executor.submit` in the class and it's the scheduler's only exit path that isn't signalled by `_stop_event`.

Suggested fix — wrap the submit and clean up on failure:

```python
try:
    executor.submit(_predict_in_process, task, str(path)).add_done_callback(_on_done)
except Exception:  # noqa: BLE001 - broken pool must not silently kill scheduler
    _log.exception('predict submit failed for %s; pool may be broken', key)
    with self._lock:
        self._in_flight.discard(key)
        self._ipc_refs[path] = max(0, self._ipc_refs.get(path, 0) - 1)
```

This keeps the scheduler thread alive for transient failures and — crucially — makes a broken-pool situation visible in logs. If you'd prefer a harder stance (stop the loop on `BrokenProcessPool` specifically), that's fine too, as long as the failure is logged before the thread exits.

zero-bang (changes requested):
Thorough, well-motivated PR. The segfault root-cause analysis is solid, the IPC lifecycle with refcounted cleanup is carefully designed, and the circular-import fix is correct (verified: `PredictLoop` is annotation-only in `shutdown_sequencer.py` with `from __future__ import annotations` in place).

One blocking concern around scheduler resilience, and one non-blocking documentation note:

**Non-blocking: TD-087 and TD-088 now reference removed code.** Both entries in `docs/TechnicalDebt.md` (lines 1082-1102) point at `_reconstruct_pooled` in `nexus/startup/sequencer.py`, which this PR deletes. The same pooled-reconstruction pattern now lives in `warm_cache.py`. Worth updating those entries to reference the new location (or marking them partially resolved) so a future reader doesn't go hunting for a method that no longer exists.

nexus/strategy/predict_loop.py
blahmonkey: Addressed in 55ad530 (`nexus/strategy/predict_loop.py:_submit_one`).

`executor.submit(_predict_in_process, task, str(path)).add_done_callback(_on_done)` is now wrapped in `try/except Exception`. On any submit-time failure (including `BrokenProcessPool`) the failure is logged via `_log.exception('predict submit failed for %s; pool may be broken', key)`, then under `self._lock` the `_in_flight` mark is discarded and the already-incremented `_ipc_refs[path]` counter is decremented with `max(0, ...)`. The scheduler thread keeps polling — a transient failure self-recovers and a persistent broken-pool situation surfaces in logs each cycle instead of dying silently after one.

**Class sweep** (every other external-IO call in `PredictLoop` audited as part of this fix):

- `_handle_predict_result` → `path.unlink(missing_ok=True)` — already wrapped in `try/except OSError` in commit `8edc904`. Not a variant.
- `tick_once` → `market_data_provider(kline_size)` — propagates by design (the synchronous backtest-replay contract; pinned by `test_market_data_provider_exception_propagates`). Not a variant.
- `_submit_due` → `market_data_provider(kline_size)` — already inside an outer `try/except` that reschedules the group. Not a variant.
- `stop()` → `executor.shutdown(...)` — shutdown path. Not a variant.
- `_write_market_data_ipc` → `market_data.write_ipc(path)` — **same class of defect**: `_ipc_current[kline_size]` is updated under the prior lock to point at a path that does not exist yet; an `OSError` on write would leave a phantom path that the next cache-hit cycle would hand to workers. **Fixed in the same commit:** `write_ipc` is now wrapped in `try/except OSError`; on failure, `_ipc_current[kline_size]` is rolled back to its prior `(signature, path)` entry (or popped if there was none) and the phantom `_ipc_refs[path]` zero-entry is popped, then the exception re-raises so `_submit_due`'s outer except catches and reschedules the group.

**Tests** (`tests/test_predict_loop.py::TestPredictLoopResilience`): `_BrokenExecutor` whose `submit` raises proves the scheduler stays alive across retries and `_ipc_refs` rolls back to 0; an `OSError`-injected `write_ipc` proves `_ipc_current` is restored and no phantom `_ipc_refs` entry remains. ruff clean, mypy clean, 58 predict-suite tests pass.

blahmonkey:
Re the non-blocking TD-087 / TD-088 note in the review body: addressed in 55ad530.

Both entries in `docs/TechnicalDebt.md` now retarget their **Module** lines and body references from `nexus/startup/sequencer.py` `_reconstruct_pooled` (deleted in v0.52.1) to `nexus/startup/warm_cache.py` `warm_cache`, where the same pooled-reconstruction pattern (`init_worker` + `data_by_dir` broadcast) now lives. The defect classes still apply unchanged — TD-087's diagnostic ambiguity (bare `KeyError` from an unseeded `_worker_data` dir indistinguishable from a per-permutation `ReconstructionError`) and TD-088's `O(num_dirs × num_workers)` memory broadcast both carry over verbatim to `warm_cache.warm_cache`. Each entry now notes the v0.52.1 move so a future reader sees the historical context.

Sweep grep across `docs/` confirmed no other stale references to `_reconstruct_pooled` or `_sensor_wire_max_workers`.

nexus/strategy/predict_loop.py
```diff
@@ -308,45 +534,230 @@ def tick_once(self, wired: WiredSensor) -> None:
...
+            _log.exception(
+                'predict submit failed for %s; pool may be broken',
+                key,
+            )
+
             with self._lock:
-                if not self._running:
-                    return
+                self._in_flight.discard(key)
+                self._ipc_refs[path] = max(0, self._ipc_refs.get(path, 0) - 1)
```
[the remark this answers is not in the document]
blahmonkey: Addressed in 1f06fcf.

Inside the `_submit_one` except branch's `with self._lock:` block, `_next_due[key] = time.monotonic() + wired.interval_seconds` is now set (guarded by `if self._running:`). This matches `_handle_predict_result`'s finally cleanup exactly (which already does discard + decrement-ipc-ref + advance-next-due) and the `_reschedule` helper (which does discard + advance-next-due). The ipc-ref decrement stays in place since `_reschedule` does not own that bookkeeping.

**Class sweep** — every `_in_flight.discard` site audited:

- `_reschedule(key)` — discards + advances. Reference implementation.
- `_handle_predict_result` finally — discards + decrements ref + advances. Correct.
- `_submit_due` failure branches (kline-size extraction, provider failure, empty market data) — all call `_reschedule(key)`. Correct.
- `_submit_one` early-exit when not running / executor gone — discards only; intentional shutdown path. Not a variant.
- `stop()` — clears `_in_flight` entirely. Not a variant.
- `_submit_one` except (this commit) — now matches `_handle_predict_result`'s finally semantics. **Fixed.**

No other variants.

**Test** (`tests/test_predict_loop.py::TestPredictLoopResilience::test_broken_pool_does_not_tight_loop_after_failure`): with `interval_seconds=1` and a `_BrokenExecutor` whose `submit` always raises, asserts that after the first failure no more than 1 additional submit attempt happens within 0.5 s. With the bug the next ~0.25 s poll would re-submit unconditionally (`after_count - first_count` would be ~2). ruff clean, mypy clean, 59 predict-suite tests pass.

zero-bang (approved):

PR #77, opened 2026-05-31
feat: SnapshotScheduler + MtmLoop; per-strategy unrealized PnL (v0.54.0)

### LINKED ISSUE #76: State checkpoint + position MTM gaps surfaced on prod (snapshot scheduler missing; Position.unrealized_pnl never marked)
Two structural state-management defects surfaced while diagnosing v0.53.1 on prod (Praxis `praxis:0.68.2` deployment, epoch 9, ~13h uptime). Both pre-date v0.53.x; neither is a regression. Both materially corrupt every state-based number the system reports.

---

### Observation

`state_store.checkpoint()` has exactly **two callers** in Nexus:

| Caller | When | Conditional? |
|---|---|---|
| [`nexus/startup/sequencer.py:519`](nexus/startup/sequencer.py) | boot-time capital reconciliation | only if `old_notional != praxis_total_notional` |
| [`nexus/startup/shutdown_sequencer.py:943`](nexus/startup/shutdown_sequencer.py) | graceful shutdown | always |

Grep confirms zero matches for `snapshot_interval`, `take_snapshot`, `SnapshotScheduler`, `periodic.*snapshot` anywhere in `nexus/`. No timer or event-count threshold triggers a checkpoint during steady-state operation.

### Live evidence (epoch 9, 2026-05-31T03:46Z observation)

| | |
|---|---:|
| `snapshot.bin` mtime | 2026-05-30T14:58:18Z (boot+6min, **13.8h stale**) |
| `snapshot.bin` size | 587,891 bytes |
| `wal.bin` mtime | 2026-05-31T03:46:29Z (actively writing) |
| `wal.bin` size | **11.5 GB** |
| Growth rate | ~1 GB/h |
| TradeClosed events since snapshot | 1,512 |

### Consequences

1. **Unbounded disk growth** — at ~1 GB/h, a week of uptime fills 168 GB on a shared volume.
2. **Every read of `snapshot.bin` between restarts is stale.** Concrete divergence we measured by running `StateStore.recover()` in a sidecar against the live files:

| Field | snapshot.bin (stale) | recover() (live) |
|---|---:|---:|
| `cumulative_realized_pnl` | +$22,744.59 | +$25,068.51 |
| `equity_hwm` | $22,744.59 | $25,126.50 |
| `total_drawdown` | $0.00 | $57.99 (0.0023%) |
| `len(state.positions)` | 1 | **4,162** |
| `position_notional` | $1,999.80 | $787,810.12 |
| `working_order_notional` | $3.3M | $7.5M |
| strategies in `risk.per_strategy` | 2,320 | 2,672 |

3. **Catastrophic restart-replay time.** Sidecar measured: **275s (4.6 min) to replay 11.5 GB**, peak RSS 12 GB. A crash recovery on this state would take ~5 min + the full wiring cycle before signals resume. The cost scales linearly with `now - last_snapshot`.

### Suggested fix

Add a `SnapshotScheduler` patterned on the existing `CacheScheduler` in Praxis. Trigger checkpoint on either:
- Wall-clock interval (e.g. every 5 min)
- Or N WAL appends since last checkpoint (e.g. 1000)
- Or both, whichever fires first

The checkpoint already acquires `wal_lock` (see `state_store.py:75-90` docstring), so concurrency with `append_mutation` / `append_event` is correctly handled.

Backwards-compatible default: scheduler off, must be opted into via env var. Easy to roll out, easy to revert.

---

### Observation

`Position.unrealized_pnl` is set on entry (typically to 0) and never re-touched until close. Aggregating across the 4,161 open positions in the live `recover()`'d state:

| | |
|---|---:|
| open positions | 4,161 |
| positions with `unrealized_pnl != 0` | **0** |
| `state.risk.unrealized_pnl` (aggregate) | 0 |
| sum of `position.size * position.entry_price` | $785,813 |
| **real MTM** (computed against binsim live mid 74185.88, longs only) | **+$3,066.35 (+0.39% of cost basis)** |

### Consequences

1. `state.risk.equity = realized + unrealized` understates real equity by the unrealized PnL of the open book. With $787K open on a $25K realized base, the open exposure is **31× the realized PnL**.
2. `total_drawdown` / `equity_hwm` derivatives all run blind to open-book P&L. A 5% adverse BTC move would carry ~$40K paper loss that the system would have zero visibility into until each position closed individually.
3. Per-strategy risk gates (`rolling_loss_24h` / `7d` / `30d`) operate on realized-only data, so a strategy carrying a large open losing position keeps full trading allowance until it actually exits.

### Suggested fix

Periodic MTM loop (could share the snapshot scheduler's timer) that:
1. Reads the latest mid/last from a `mark_price_provider` callable (Praxis already has the market-data cache providing this for `produce_signal`).
2. For each `state.positions[pid]`: `position.unrealized_pnl = (mark - entry) * size * side_sign`
3. Sums into `state.risk.unrealized_pnl`
4. Triggers `state.risk.update_cumulative_realized_pnl(...)` equivalent for unrealized so drawdown derivatives stay coherent

Same lock chain as `serialize_state` snapshotting; same `positions_lock` discipline.

Cadence can be loose (e.g. every 30s) — risk gates don't need tick-level MTM.

---

## Repo scope

Both fixes live entirely in Nexus (`nexus/infrastructure/state_store.py`, possibly a new `nexus/infrastructure/snapshot_scheduler.py` and `nexus/infrastructure/mtm_loop.py`, plus wiring in `nexus/startup/sequencer.py`). No Praxis-side change required for either.

## Priority

- **#1 is operational** — affects disk, restart latency, snapshot reliability. Becomes more painful the longer the box runs. Recommend prioritising.
- **#2 is correctness** — affects every risk-gate decision involving an open position. Subtle but cumulative.

Neither is a v0.53.x regression.

blahmonkey:

Closes #76.

## What does this PR change?

Adds two periodic schedulers that close the operational + risk-visibility gaps surfaced on the v0.68.x Praxis prod deployment (Nexus #76, Praxis #127). Pure additive surface — no existing call paths change. Praxis-side wiring (instantiating both in `praxis/launcher.py` next to `HealthLoop`) ships in a separate Praxis PR.

### `feat:` SnapshotScheduler — wall-clock periodic `state_store.checkpoint(state)`

Patterned on [`HealthLoop`](nexus/core/health_loop.py): daemon `threading.Timer`, `start`/`stop`/`tick_once`, idempotent, lock-safe. Holds `positions_lock` + `CapitalController._lock` around the checkpoint call so the snapshot captures a consistent point-in-time view; `state_store.checkpoint()` acquires `wal_lock` internally — same lock pattern as [`shutdown_sequencer._checkpoint_state`](nexus/startup/shutdown_sequencer.py). Checkpoint failures (disk full, transient I/O, msgpack failure) are caught + logged at ERROR; the loop continues. Default interval 300 s.

The scheduler's internal `_lock` is intentionally NOT acquired inside the heavy lock chain — that would add a new lock between `CapitalController._lock` and `wal_lock` in the documented ordering. A racing `stop()` between the `_tick` entry check and the checkpoint completing simply writes one extra snapshot; duplicate checkpoints are harmless (no ratchet semantics to defeat).

### `feat:` MtmLoop — wall-clock periodic mark-to-market of `state.positions`

Iterates the open positions under `positions_lock`, asks a `mark_price_provider` callable for each symbol's mark, computes per-position `(mark - entry) * size * side_sign`, writes back to `Position.unrealized_pnl`, **buckets per `strategy_id`** into `StrategyRiskState.strategy_unrealized_pnl` (creating `StrategyRiskState` lazily for strategies that opened a position but never exited — OutcomeProcessor creates them on first exit), and writes the instance aggregate via `RiskState.update_unrealized_pnl` which calls `recompute_drawdown_metrics` so `equity` / `equity_hwm` / `total_drawdown` / `unrealized_drawdown` stay coherent.

If the provider returns `None` or a non-finite mark for any symbol, the entire tick is aborted without partial writes — stale marks are preferred over half-marked snapshots. Provider exceptions are caught + logged at ERROR via `_log.exception`; loop continues. Default interval 30 s.

### `feat:` `strategy_unrealized_pnl` field on `StrategyRiskState`

Signed `Decimal`, finite-checked in `__post_init__`. [`_encode_strategy_risk_state`](nexus/infrastructure/wal_codec.py) emits the new key; [`_decode_strategy_risk_state`](nexus/infrastructure/wal_codec.py) defaults missing values to `'0'` so snapshots / WAL entries written by older Nexus versions (≤ v0.53.x) decode without error and the MtmLoop overwrites the zero on the next tick. Backward-compat with no migration step.

## Why this PR exists — observed evidence from prod

On `praxis:0.68.2` (epoch 9), live recovery against the live snapshot + WAL revealed:

| Field | snapshot.bin (stale at boot+6 min) | recover() (true current) |
|---|---:|---:|
| `cumulative_realized_pnl` | +$22,744.59 | +$25,068.51 |
| `len(state.positions)` | 1 | **4,162** |
| `position_notional` | $1,999.80 | $787,810.12 |
| `unrealized_pnl` (aggregate, never marked) | $0.00 | **$0.00** ← all 4,161 positions had `unrealized_pnl=0` |
| Real MTM @ binsim mid (computed externally) | n/a | **+$3,066.35** |

Without #1: WAL grew unbounded (~1 GB/h, 11.5 GB observed at one point), restart-replay took 275 s with 12 GB peak RSS, every read of `snapshot.bin` between restarts was 13+ hours stale.
Without #2: every risk-gate decision involving an open position ran blind to open-book P&L. A 5 % adverse move would have hidden ~$40 K of paper loss from risk gates until each position closed individually.

## Verification

- `ruff check` on touched files — clean
- `mypy --strict` on `snapshot_scheduler.py` + `mtm_loop.py` — clean
- `pytest tests/test_snapshot_scheduler.py tests/test_mtm_loop.py -v` — **29 passed** (8 SnapshotScheduler + 21 MtmLoop)
- `pytest tests/` — **1438 passed, 17 skipped** (was 1409 baseline, +29 from new test surface)
- Backward-compat decode of `strategy_unrealized_pnl` missing key — implicitly tested by the existing 1409 tests still loading their fixtures cleanly
- pr-prep Phase 1b LLM-attribution: PASS
- pr-prep Phase 4 Greybeard (final): **ok**

## Scope

`nexus/infrastructure/snapshot_scheduler.py` (new), `nexus/core/mtm_loop.py` (new), `nexus/core/domain/risk_state.py` (new field + finite-check), `nexus/infrastructure/wal_codec.py` (encode new field + backward-compat decode), `tests/test_snapshot_scheduler.py` (new, 8 cases), `tests/test_mtm_loop.py` (new, 21 cases incl. per-strategy attribution + lazy `StrategyRiskState` creation + no-positions drawdown recompute), `pyproject.toml` (`0.53.1 → 0.54.0`), `CHANGELOG.md` (v0.54.0).

Praxis-side wiring is **not** in this PR — `SnapshotScheduler` and `MtmLoop` need to be instantiated in `praxis/launcher.py` alongside `HealthLoop` (passing `positions_lock`, `capital_controller.lock_cm`, and a `mark_price_provider` from the market-data cache). Shipped as a follow-up Praxis PR once this lands.

## Checklist

- [x] I ran `python -m pytest` locally without errors (1438 passing, 17 skipped)
- [x] I updated CHANGELOG.md


nexus/core/mtm_loop.py
```diff
@@ -0,0 +1,244 @@
...
+            or not isinstance(interval_seconds, (int, float))
+            or interval_seconds <= 0
+        ):
+            msg = 'MtmLoop.interval_seconds must be a positive number'
+            raise ValueError(msg)
+
+        self._state = state
+        self._mark_price_provider = mark_price_provider
+        self._interval_seconds = float(interval_seconds)
+        self._positions_lock = positions_lock
```
zero-bang, line 112: **Blocking: the lock-identity invariant is assumed but not enforced.** `_mark()` mutates `state.risk.per_strategy` (creates `StrategyRiskState`, writes `strategy_unrealized_pnl`) and calls `state.risk.update_unrealized_pnl()` (which writes `equity`, `equity_hwm`, `total_drawdown`, `max_drawdown`, …) while holding only `positions_lock`. Those exact structures are otherwise guarded by `state.risk.lock`: OutcomeProcessor writes them under `state.risk.lock_cm()` (`outcome_processor.py:553`, `:222`), and the validator hot path reads them under the same lock every action (`risk_stage.py:130` → `to_risk_check_metrics`). Correctness therefore depends entirely on the launcher wiring `state.risk.lock is positions_lock` (same object). If a future Praxis wiring passes any other lock, MtmLoop and OutcomeProcessor mutate `per_strategy` concurrently with no mutual exclusion — re-opening the FINAL-MAJOR-02 `RuntimeError: dictionary changed size during iteration` / lost-update race that the codebase invested heavily to close. The Praxis-side wiring ships in a *separate* PR, so nothing in this PR guards the assumption.

`ShutdownSequencer.__init__` already treats this exact invariant as fatal (`shutdown_sequencer.py:188-204`, raises `RuntimeError` when `state.risk.lock is not positions_lock`). Add the same guard here:

```python
if positions_lock is not None and (
    not hasattr(state.risk, 'lock') or state.risk.lock is not positions_lock
):
    msg = (
        'MtmLoop requires `state.risk.lock is positions_lock` whenever '
        '`positions_lock` is supplied, so a single acquisition serializes '
        'MtmLoop`s per_strategy / RiskState writes against OutcomeProcessor '
        'and the validator reader.'
    )
    raise RuntimeError(msg)
```

nexus/infrastructure/snapshot_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+            or not isinstance(interval_seconds, (int, float))
+            or interval_seconds <= 0
+        ):
+            msg = 'SnapshotScheduler.interval_seconds must be a positive number'
+            raise ValueError(msg)
+
+        self._state_store = state_store
+        self._state = state
+        self._interval_seconds = float(interval_seconds)
+        self._positions_lock = positions_lock
```
zero-bang, line 113: Same unenforced-invariant gap as `MtmLoop` (lower severity here since this path only reads). `checkpoint()` → `serialize_state()` iterates `state.risk.per_strategy` (and `state.capital.per_strategy_deployed`), but this scheduler holds only `positions_lock` + `capital_lock_cm` around it. Consistency with the rest of the snapshot path relies on `state.risk.lock is positions_lock`. `ShutdownSequencer` — whose `_checkpoint_state` this is explicitly patterned on — enforces that identity at construction (`shutdown_sequencer.py:188-204`); this scheduler accepts the lock without checking. Add the same `state.risk.lock is positions_lock` assertion so a mis-wired launcher fails loudly at construction rather than producing an unguarded `per_strategy` iteration against a live OutcomeProcessor insert.

nexus/infrastructure/snapshot_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+        positions_cm: AbstractContextManager[bool | None] = (
+            self._positions_lock if self._positions_lock is not None else nullcontext()
+        )
+        capital_cm: AbstractContextManager[bool | None] = (
+            self._capital_lock_cm() if self._capital_lock_cm is not None else nullcontext()
+        )
+
+        try:
+            with positions_cm, capital_cm:
+                self._state_store.checkpoint(self._state)
```
zero-bang, line 208: Non-blocking tradeoff worth surfacing: `checkpoint()` serializes the full state **and** does the snapshot disk write + fsync + WAL truncate, all under `_wal_lock` (`state_store.py:75-90`), while this scheduler holds `positions_lock` (== `state.risk.lock`, the validator's lock) for the entire call. Every 300s the per-action validator path (`risk_stage.py:130` acquires `state.risk.lock`) stalls for the full serialize+fsync duration. This is the TD-073 latency-spike class, but now recurring during live trading rather than only at shutdown — and the prod evidence in this PR (4,162 open positions) is exactly the state size where serialization is non-trivial. Either restructure so only serialization holds `positions_lock` and the fsync happens outside it, or (minimum) record this as a Technical Debt entry referencing TD-073 so the periodic-checkpoint latency cost is tracked rather than silently shipped.

nexus/core/mtm_loop.py
```diff
@@ -0,0 +1,244 @@
...
+                        self._state.risk.per_strategy[strategy_id] = new_srs
+                        existing = new_srs
+
+                    existing.strategy_unrealized_pnl = unrealized
+
+                for strategy_id, srs in self._state.risk.per_strategy.items():
+                    if strategy_id not in per_strategy and srs.strategy_unrealized_pnl != _ZERO:
+                        srs.strategy_unrealized_pnl = _ZERO
+
+                self._state.risk.update_unrealized_pnl(aggregate)
```
zero-bang, line 309: Behavior change worth confirming is intended: `update_unrealized_pnl` → `recompute_drawdown_metrics` ratchets `max_drawdown` / `max_drawdown_pct` (`risk_state.py:234-235`, lifetime `max(...)`), and `total_drawdown` now folds in `unrealized_pnl` via `equity`. With MtmLoop marking every 30s, a *transient* adverse mark that fully recovers still permanently raises `max_drawdown` / `max_drawdown_pct` — and those feed the validator's max-drawdown gate (`to_risk_check_metrics`). Pre-PR, unrealized was never marked, so `max_drawdown` only captured realized swings. Forward-looking failure mode: in a volatile session a brief mark spike on the open book latches `max_drawdown_pct` high and keeps a max-drawdown gate tripped long after price recovers, even with zero realized loss. If that's the desired semantics (mark-to-market drawdown), fine — but confirm the max-drawdown gate threshold was calibrated against realized-only drawdown and won't false-trip once unrealized is folded in.

zero-bang (changes requested):
Solid, well-documented additive surface — the `HealthLoop` patterning is faithful, the abort-on-bad-mark discipline in `MtmLoop` is the right call, the backward-compat decode of `strategy_unrealized_pnl` is correct, and the per-strategy attribution + lazy-create tests are thorough. One blocking concern: both new components mutate/read `state.risk` structures under `positions_lock` only, depending on the unenforced launcher invariant `state.risk.lock is positions_lock`. `ShutdownSequencer` hard-fails when that invariant is violated; these don't, and the Praxis wiring that establishes it ships in a separate PR — so a misconfiguration re-opens the FINAL-MAJOR-02 race silently. Details inline, plus two non-blocking notes (checkpoint-latency tradeoff and a max-drawdown ratchet behavior change worth confirming).

nexus/core/mtm_loop.py
blahmonkey: Addressed in 30f57f1.

`MtmLoop.__init__` now asserts `state.risk.lock is positions_lock` whenever `positions_lock` is supplied, with the same message shape as `shutdown_sequencer.py:188-204`. A mis-wired launcher (any lock other than `state.risk.lock`) raises `RuntimeError` at construction rather than entering steady-state with the lock-identity invariant silently violated.

Class sweep: the same gap exists on `SnapshotScheduler`. Addressed in the same commit (see thread on `snapshot_scheduler.py:79`). PLUS the additional P2 finding — `SnapshotScheduler` now also requires `capital_lock_cm` when `positions_lock` is supplied (mirrors `shutdown_sequencer.py:206-216`), so a partial lock-cluster wiring fails loudly too.

Tests: 3 cases per scheduler (mismatch rejected, match accepted, no-positions-lock path accepts any `risk.lock`) plus a SnapshotScheduler-specific case for the `capital_lock_cm` requirement. Pre-existing tests that supplied `positions_lock` without setting `state.risk.lock` also got the wiring fix.

nexus/infrastructure/snapshot_scheduler.py
blahmonkey: Addressed in 30f57f1. `SnapshotScheduler.__init__` now asserts both:

1. `state.risk.lock is positions_lock` whenever `positions_lock` is supplied — same message shape as `shutdown_sequencer.py:188-204`.
2. `capital_lock_cm` is required whenever `positions_lock` is supplied — same shape as `shutdown_sequencer.py:206-216`. The positions_lock + capital_lock_cm form a lock-cluster; partial wiring leaves the capital-side `dictionary changed size during iteration` race against OutcomeProcessor reachable.

This addresses both your original thread and the P2 follow-up about partial lock-cluster wiring. Tests in `tests/test_snapshot_scheduler.py`: `test_lock_identity_mismatch_rejected_at_construction`, `test_lock_identity_match_accepted`, `test_no_positions_lock_accepts_any_risk_lock`, `test_capital_lock_cm_required_when_positions_lock_supplied`.

nexus/infrastructure/snapshot_scheduler.py
blahmonkey: Recorded as TD-091 in 30f57f1 — `docs/TechnicalDebt.md` now carries an entry explicitly cross-referencing TD-073 and capturing the recurrence-during-live-trading angle plus the migration path.

The migration: split `save_snapshot(state, path, wal)` into `serialize(state) -> bytes` (under positions_lock + capital_lock_cm) + `persist(payload, path, wal)` (lock-free, fsync-bound, wal_lock-only). The same atomicity guarantee is preserved because wal_lock (innermost) still wraps the persist+truncate pair so concurrent `append_mutation` cannot interleave. Restructuring `state_store` internals is scope creep into a load-bearing path that would have benefited from its own PR + cross-test against the existing crash-recovery contracts; going with your minimum-acceptable path (TD entry) and deferring the restructure until either a validator-latency spike correlates with the checkpoint cadence on prod OR a feature lands that meaningfully increases per-checkpoint serialize size.

Tunable mitigation noted in the TD entry: raise `NEXUS_SNAPSHOT_INTERVAL_SECONDS` above the 300s default to trade WAL replay-on-recovery time for fewer validator stalls per hour.

nexus/core/mtm_loop.py
blahmonkey: Sharp catch — confirmed the ratchet in `risk_state.py:241-242` and the validator-gate-trip risk. NOT the desired semantics. Fixed by splitting in 30f57f1:

- `max_drawdown` / `max_drawdown_pct` — now ratchet from **realized-only** drawdown (was: total drawdown). Validator gate threshold stays valid; pre-existing calibration unchanged. A transient adverse mark that fully recovers no longer permanently raises the lifetime peak.
- `max_total_drawdown` / `max_total_drawdown_pct` — **new fields** that ratchet from total (realized + unrealized) drawdown. Exposed for telemetry and for any future gate that wants strictly tighter mark-aware discipline.

Backward compat: pre-v0.54.0 `unrealized_pnl` stayed at zero, so `total_drawdown == realized_drawdown` always, so existing `max_drawdown` values in snapshots are mechanically equivalent to a realized-only ratchet. The field interpretation has been narrowed from "total" to "realized-only" without changing any historical number. `_encode_risk_state` emits the new keys; `_decode_risk_state` defaults missing keys to `0` so older snapshots decode without error and the next MtmLoop tick bumps the new field if needed.

Tests: `test_recompute_drawdown_metrics_tracks_losses_without_new_peak` and `test_recompute_drawdown_metrics_preserves_lifetime_max_drawdown` rewritten to pin the split. The OLD assertions encoded the very behaviour you flagged as questionable; the new assertions verify the validator-stable realized-only ratchet AND the new total-equity ratchet on separate fields.

Class sweep: this is the only call site that ratchets max-drawdown. `MtmLoop._mark` -> `update_unrealized_pnl` -> `recompute_drawdown_metrics` is the chain; no other ratchet writers exist. Validator reads `max_drawdown_pct` (per `risk_stage.py:130` -> `to_risk_check_metrics`), which retains realized-only semantics.

blahmonkey:
Two round-2 findings from the P2 review (forwarded by maintainer) — addressed in 7418dfc.

**P2 deadlock — MtmLoop provider call held positions_lock.**

Verified: the lock is non-reentrant, the provider could re-enter `state.risk.lock_cm()` and hang. Restructured `_mark()` into three phases per your suggested shape:

- **Phase A** snapshots the set of held symbols under `positions_lock`; releases.
- **Phase B** fetches marks lock-free. Provider can safely call back into risk APIs.
- **Phase C** re-acquires the lock, validates no NEW symbol opened during Phase B (aborts if so — we have no mark for it), recomputes from the CURRENT `Position.size`/`entry_price` (so partial-fill mutations between phases are respected), writes per-position + per-strategy + aggregate.

Positions closed during Phase B are simply absent from Phase C's iteration. The mutation-discipline invariant (positions can't be added or removed mid-write) holds because all writes happen under Phase C's lock.

New test `test_provider_can_reacquire_risk_lock_without_deadlock` wires a provider that acquires `state.risk.lock` mid-call and asserts `tick_once()` completes within 3s. Existing `test_tick_holds_positions_lock` renamed to `test_provider_called_without_positions_lock` and the assertion flipped to pin the new (correct) contract.

**P2 per-strategy unrealized PnL not consumed by current risk gates.**

Right. `RiskCheckMetrics` (`risk_state.py:64-73`) exposes instance-level fields only; `risk_stage.py:66` reads `total_drawdown` / `max_drawdown` / rolling losses, none of which carry per-strategy unrealized. My CHANGELOG and MtmLoop module docstring overclaimed.

Softened both to match reality: the per-strategy field is persisted, written every MtmLoop tick, visible to telemetry / `recover()` / external readers, and ratchets correctly — but is NOT yet read by the validator's per-action gates. Wiring it through (extending `RiskCheckMetrics` with a per-strategy field and updating `risk_stage` to consult it) is a separate scope and a separate gate-calibration discussion.

Full suite: 1446 passed, 17 skipped.

zero-bang (approved):

PR #83, opened 2026-06-09
fix: account_dust ledger + close_as_dust for sub-lot EXIT residue (v0.57.0)

### LINKED ISSUE #82: fix: persistent dust positions from sub-lot residue + intake-rejected dust EXITs (depends on Vaquum/Praxis#142)
## Problem

Production observation (Praxis v0.75.0, epoch=16, ~3h post-deploy): **1,325 stuck trades** in the active strategy account emit EXIT signals every 15–30 min for sub-lot-min positions. Each is rejected at the Praxis launcher's intake quantization (`praxis/launcher.py:772-783`, `INTAKE_BELOW_MIN_QTY`), the EXIT command is silently dropped, the position stays alive forever.

Root cause (verified via Praxis spine traces + Nexus source read; see codex-supervised audit findings):

- Strategy emits `Action(action_type=EXIT, size=position.size - position.pending_exit)` — full position close.
- Praxis launcher quantizes `action.size` to `LOT_SIZE.stepSize` (e.g. `0.03194842` → `0.03194000` for BTCUSDT `stepSize=0.00001`).
- Quantized qty submitted to venue, fill comes back, `OutcomeProcessor._reduce_position` (`outcome_processor.py:471-518`) decrements `position.size` by the quantized qty.
- Residue = original full-precision qty − quantized qty (e.g. `0.00000842` BTC).
- `Position.is_closed` is exact-zero (`nexus/core/domain/position.py:91-94`, `return self.size == _ZERO`), so the residue keeps the trade alive in `state.positions`.
- Next tick: strategy emits EXIT with `size=residue` (sub-lot). Praxis launcher's `quantize_for_command` rejects it as below `lot_min`, `_build_exit_context` returns `None`, no `OrderContext`, no `pending_exit`, no state change. Infinite retry loop.

Two blocker paths flow through the same launcher function: **B1** (post-fill residue) and **B2** (intake-rejected sub-lot EXIT). Single fix point covers both.

## Fix shape (codex-supervised audit, AUDIT_SIGNED_OFF)

Nexus-side scope of this issue. Praxis-side scope tracked in `Vaquum/Praxis#142`.

### Schema additions

- `nexus/core/validator/pipeline_models.py` — add `ValidationRequestContext.intended_full_close: bool = False`.
- `nexus/infrastructure/praxis_connector/order_context.py` — add `OrderContext.intended_full_close: bool = False`.
  - **Both** dataclasses need the field because `_build_order_context` (`praxis/launcher.py:913`) rebuilds the validation context by re-running `build_context` rather than passing the original through. A field only on `OrderContext` would be lost on the rebuild.
- `nexus/core/domain/instance_state.py` — add `account_dust: dict[str, Decimal] = field(default_factory=dict)`. Symbol-keyed. Durable.
- `nexus/infrastructure/wal_codec.py` — encode/decode `account_dust` in WAL/snapshot. Default to `{}` for old snapshots.

### Behavior changes

- `OutcomeProcessor._reduce_position` — after the terminal EXIT fill decrement, if `context.intended_full_close and position.size > _ZERO`, move the residue into `state.account_dust[position.symbol]`, zero `position.size`, then proceed to the existing `if position.is_closed: del state.positions[trade_id]` branch. Covers B1.
- `OutcomeProcessor.close_as_dust(trade_id, reason, command_id_or_dust_id)` — new method called from Praxis launcher when intake quantization rejects a full-close EXIT as sub-lot. Removes the position from `state.positions`, adds `position.size` to `state.account_dust[symbol]`. Covers B2.

### Constraints

- **Idempotency (M1)**: `close_as_dust` must be safe under WAL replay and command retries. Use deterministic dedup key — `(trade_id, command_id)` or a synthetic `dust_close_id = f'dust-{command_id}'`.
- **Durability (M2)**: `account_dust` participates in WAL replay + snapshot save/load. Old snapshots without the field default to `{}`.
- **No fake execution facts (M3)**: do NOT emit `TradeClosed` / `TradeOutcomeProduced` for the intake-dust path — those are execution facts owned by Praxis `ExecutionManager` after a real venue fill. For dust closure, emit a distinct Nexus state-transition log/audit fact if observability is needed.

### Equality semantics

`intended_full_close` is set at the Praxis launcher boundary as `action.size == (position.size - position.pending_exit)` (exact Decimal equality). The current prod strategy `examples/strategies/logreg_binary_evsfd._exit` always emits `size=remaining`, so this evaluates True every time for the active strategy family. Future strategies wanting fuzzy "close enough" semantics should add an explicit `Action` field instead of relaxing this comparison.

## Out of scope

- Reconciliation of the existing 1,325 stuck positions — the operator will reset epoch to wipe them.
- Symbol-filter plumbing into Nexus (full "C-pure" shape) — not needed; Praxis already owns the lot_step truth via `binance_adapter.quantize_for_command` and carries `intended_full_close` across the boundary.

## Tests

- Unit: `_reduce_position` with `intended_full_close=True` zeroes residue + populates `account_dust[symbol]`.
- Unit: `close_as_dust` idempotency under repeated calls with the same dedup key.
- Unit: `OrderContext` carries `intended_full_close` from `ValidationRequestContext` via `_build_order_context` (this is the regression test for the codex-identified rebuild-loses-flag issue).
- WAL/snapshot: `account_dust` roundtrips. Old snapshots without the field load cleanly.
- E2E equivalent: full ENTER → snapped EXIT cycle leaves position closed and `account_dust` populated with the residue.

## References

- Codex-supervised audit final findings: `AUDIT_SIGNED_OFF` after 2 sweep cycles. Recommended fix shape: "C-refined at the launcher boundary + terminal-fill residual close".
- Companion Praxis issue: Vaquum/Praxis#142 (launcher.py changes).

blahmonkey:

Companion to Praxis [`Vaquum/Praxis#142`](https://github.com/Vaquum/Praxis/issues/142) — this PR ships the **Nexus side only**. The two are paired: Nexus alone is dormant (the new `intended_full_close` field defaults to `False` so behavior is unchanged on every existing caller), but Praxis cannot complete the fix without these schema and behavior changes here.

Closes [#82](https://github.com/Vaquum/Nexus/issues/82).

## What does this PR change?

Adds the Nexus-side schema and behavior to close sub-lot dust positions cleanly. After this PR + the Praxis companion, full-position EXIT actions whose snapped quantity leaves residue (or whose snapped quantity falls entirely below `lot_min`) route the residue into a per-symbol `account_dust` ledger on `InstanceState` instead of leaving the position alive forever in `state.positions`.

### `feat:` `intended_full_close: bool` on `ValidationRequestContext` and `OrderContext`

New field on [`ValidationRequestContext`](nexus/core/validator/pipeline_models.py) and [`OrderContext`](nexus/infrastructure/praxis_connector/order_context.py), both default `False`. Set by the Praxis launcher's `_build_exit_context` when `Action.size == position.size - position.pending_exit` (exact `Decimal` equality at intake time). `__post_init__` validates `bool` type on both classes; `OrderContext` additionally rejects `intended_full_close=True` when `is_entry=True` since the semantics only apply to EXIT orders.

The field has to live on **both** dataclasses because `praxis/launcher.py:_build_order_context` at line 913 re-runs `build_context` to reconstruct `ValidationRequestContext` rather than passing the original through. A field only on `OrderContext` would be lost across the rebuild; a field only on `ValidationRequestContext` would not reach `OutcomeProcessor`.

### `feat:` `account_dust: dict[str, Decimal]` on `InstanceState`

New per-symbol base-asset residue ledger on [`InstanceState`](nexus/core/domain/instance_state.py), defaulting to `{}`. `__post_init__` validates every key is a non-empty string and every value is a finite non-negative `Decimal`. Populated by the two close paths below; never read by execution logic — it is an accounting bucket so the base asset that the venue holds (but Nexus cannot exit) stays auditable.

### `feat:` Post-fill residue path in `_reduce_position`

[`OutcomeProcessor._reduce_position`](nexus/infrastructure/praxis_connector/outcome_processor.py) now routes residue into `account_dust` after a terminal EXIT fill. Concretely: after the existing `position.size = position.size - fill_size` decrement, if `outcome.outcome_type == TradeOutcomeType.FILLED` AND `context.intended_full_close` AND `position.size > _ZERO`, the function moves the residue into `state.account_dust[position.symbol]`, zeroes `position.size`, then falls through to the existing `if position.is_closed: del state.positions[trade_id]` branch. Partial-exit and non-terminal-fill paths are unchanged.

### `feat:` Pre-submit dust path `OutcomeProcessor.close_as_dust`

New [`close_as_dust(trade_id, reason, dust_close_id)`](nexus/infrastructure/praxis_connector/outcome_processor.py) method called by the Praxis launcher when intake-time `quantize_for_command` rejects a full-close EXIT as sub-lot (the position is so small that even the lot-snapped exit qty is below `LOT_SIZE.minQty`). Removes the position from `state.positions`, adds `position.size` to `state.account_dust[symbol]`. No venue order ever existed — this method does NOT emit `TradeClosed` / `TradeOutcomeProduced` since those are execution facts owned by Praxis's `ExecutionManager` after a real fill.

Idempotent on `dust_close_id` via a new `_processed_dust_close_ids: set[str]` (separate from `_processed_outcome_ids` so dust-close dedup is independent of regular outcome dedup). Convention is `dust_close_id = f'dust-{command_id}'` for stable keys across WAL replay.

### `fix:` WAL/snapshot codec roundtrips `account_dust`

[`wal_codec.serialize_state`](nexus/infrastructure/wal_codec.py) encodes `account_dust` as `{symbol: str(Decimal)}` alongside the existing fields; `_decode_state_v1` reads via `d.get('account_dust', {})` so pre-existing snapshots without the field load cleanly with the field defaulted to `{}`.

## Why this PR exists — observed evidence from prod

On Praxis prod (v0.75.0, epoch=16, ~3h post-deploy on 2026-06-09), 1,325 trades were stuck in an infinite EXIT-retry loop. Concrete trace from trade `3ea13679-8f2f-4741-96d9-2e6a5f1b0e2a`:

| Time (UTC) | Event |
|---|---|
| 06:18:47 | ENTER command 3ea13679: MARKET BUY `quote_qty=$2000`, filled at $63,394.61 |
| 06:33:48 | EXIT command 7b7015b5 (trade_id=3ea13679): FillReceived side=SELL, qty=`0.03154000`, TradeClosed emitted by Praxis |
| 08:03 / 08:19 / 08:49 / continuing | Strategy `perm_0019` emits EXIT for trade 3ea13679 with `requested_size=0.00000842343852261256911273687` BTC; Praxis intake rejects `INTAKE_BELOW_MIN_QTY` every cycle |

Math: ENTER fill qty `0.03154842` BTC (full Decimal precision after fee accounting) − SELL qty `0.03154000` BTC (lot-snapped to `LOT_SIZE.stepSize=0.00001`) = `0.00000842` BTC residue. `Position.is_closed` is exact-zero (`nexus/core/domain/position.py:91-94`) so the residue kept the trade alive in `state.positions`. Strategy `_exit` emits full-position EXIT every tick (`size=position.size - position.pending_exit`); next tick's value is the sub-lot residue; Praxis launcher rejects; cycle repeats forever.

Two BLOCKER paths flow through the same `praxis/launcher.py:_build_exit_context` function — post-fill residue (B1, fixed in `_reduce_position`) and pre-submit sub-lot rejection (B2, fixed via `close_as_dust`). Single fix point.

## Verification

- `ruff check nexus tests` — clean
- `pytest tests/` — **1487 passed, 17 skipped** (was 1473 + 17; added 14 new tests)
- 14 new tests: `TestDustClose` (7 cases in `tests/test_outcome_processor.py` covering B1 residue → dust, B2 close_as_dust happy path + idempotency + missing-position + per-symbol accumulation), `intended_full_close` validation (4 cases in `tests/test_order_context.py`), `TestAccountDustRoundtrip` (3 cases in `tests/test_wal_codec.py` covering empty + populated + legacy-snapshot-defaults)
- pr-prep Phase 1b LLM-attribution: PASS
- pr-prep Phase 4 Greybeard: **mediocre → fix-both applied** (2 VALID_FIX addressed: codex reference stripped from CHANGELOG NOTE per house style, test fixture symbols normalized to `BTCUSD` to match sibling tests)

## Scope

[`nexus/core/validator/pipeline_models.py`](nexus/core/validator/pipeline_models.py) (`intended_full_close` on `ValidationRequestContext`), [`nexus/infrastructure/praxis_connector/order_context.py`](nexus/infrastructure/praxis_connector/order_context.py) (`intended_full_close` on `OrderContext` + entry-rejection guard), [`nexus/core/domain/instance_state.py`](nexus/core/domain/instance_state.py) (`account_dust` ledger + validation), [`nexus/infrastructure/praxis_connector/outcome_processor.py`](nexus/infrastructure/praxis_connector/outcome_processor.py) (`_reduce_position` residue routing + new `close_as_dust` method + `_processed_dust_close_ids` dedup set), [`nexus/infrastructure/wal_codec.py`](nexus/infrastructure/wal_codec.py) (`account_dust` encode/decode), [`tests/test_outcome_processor.py`](tests/test_outcome_processor.py) (7 new dust tests), [`tests/test_order_context.py`](tests/test_order_context.py) (4 new field tests), [`tests/test_wal_codec.py`](tests/test_wal_codec.py) (3 new roundtrip tests), [`pyproject.toml`](pyproject.toml) (`0.56.0 → 0.57.0`), [`CHANGELOG.md`](CHANGELOG.md) (v0.57.0).

## Checklist

- [x] I linked issue to auto-close on merge (e.g., “Fixes #123”) when applicable


nexus/infrastructure/praxis_connector/outcome_processor.py
```diff
@@ -514,6 +515,29 @@ def _reduce_position(
...
+            if (
+                outcome.outcome_type == TradeOutcomeType.FILLED
+                and context.intended_full_close
+                and position.size > _ZERO
+            ):
+                residue = position.size
+                self._state.account_dust[position.symbol] = (
+                    self._state.account_dust.get(position.symbol, _ZERO) + residue
+                )
+                position.size = _ZERO
```
zero-bang, line 536: Non-blocking — capital-aggregate seam worth a doc note. On a full-close fill, `order_exit` (called earlier in `_handle_fill`) decrements `position_notional`/`per_strategy_deployed` by `avg_cost_basis * fill_size`, i.e. only the *sold* qty. Entry had deployed `avg_cost_basis * (fill_size + residue)`. After this branch zeroes `position.size` and the position is deleted at L529, the residue's cost basis (`avg_cost_basis * residue`) stays in the deployed aggregates with no backing position, until the next `_reconcile_capital` rebuild from live positions corrects it. The amount is sub-lot dust (~cents) and self-corrects, so not a blocker — but since `account_dust` is explicitly base-asset-only, a one-line note here that the quote-side residue is intentionally left to reconcile would save the next reader the trace.

nexus/infrastructure/praxis_connector/outcome_processor.py
```diff
@@ -534,6 +558,104 @@ def _clear_pending_exit(self, context: OrderContext, size: Decimal) -> bool:
...
+
+                return False
+
+            residue = position.size
+            symbol = position.symbol
+            self._state.account_dust[symbol] = (
+                self._state.account_dust.get(symbol, _ZERO) + residue
+            )
+            position.size = _ZERO
+            del self._state.positions[trade_id]
```
zero-bang, line 659: Non-blocking — same capital seam, more pronounced here. `close_as_dust` removes the position with no `order_exit` at all, so the *entire* position's cost basis (`avg_cost_basis * size`) remains in `position_notional`/`per_strategy_deployed` until reconcile. Sub-lot by construction (B2 only fires when the lot-snapped exit qty is below `lot_min`), so dust-sized and self-correcting — but the contract that callers must rely on `_reconcile_capital` to true up the deployed aggregate after a dust close is currently implicit. Worth stating in the docstring alongside the existing 'never read by execution logic' note.

zero-bang (dismissed):
Reviewed the full diff and traced the change across the WAL/snapshot persistence path, the capital-accounting seam, and the Nexus/Praxis component boundary.

**Verdict: approve.** The change is correct for its stated scope, dormant by default (`intended_full_close` defaults `False`), well-tested (14 new tests covering B1/B2 + roundtrip + validation), and all 5 CI checks pass (CodeQL, Mypy, Ruff, Tests). No other reviewer threads are open.

Verified: `account_dust` rides the existing full-state `serialize_state` on STATE_MUTATION entries, so B1 residue is durable on the established `_reduce_position` persistence path; the legacy-snapshot default (`d.get('account_dust', {})`) is correct; `close_as_dust` dedup is concurrency-safe because `_positions_cm` serializes and position deletion is the real double-credit guard; idempotency across restart is anchored on persisted position absence (each retry tick carries a fresh `command_id`), not the in-memory dedup set.

Two non-blocking observations inline about the capital-aggregate seam and the unbounded-residue assumption — neither blocks merge given dust-sized amounts and reconcile self-correction, but both are worth a sentence in the docstrings for the next engineer.

Forward-looking failure mode: the dust path trusts that a terminal `FILLED` on an `intended_full_close` order only ever leaves sub-lot residue. The first time the venue/Praxis emits a `FILLED` whose `fill_size` is materially short of the ordered qty (not just lot-snapped), a real base-asset remainder gets silently moved to `account_dust` and the position deleted. Worth keeping in mind when the Praxis companion lands.

nexus/infrastructure/praxis_connector/outcome_processor.py
```diff
@@ -534,6 +558,104 @@ def _clear_pending_exit(self, context: OrderContext, size: Decimal) -> bool:
...
+
+        with self._dedup_lock:
+
+            if dust_close_id in self._processed_dust_close_ids:
+                _log.debug(
+                    'duplicate dust close dropped: dust_close_id=%s trade_id=%s',
+                    dust_close_id,
+                    trade_id,
+                )
+                return False
```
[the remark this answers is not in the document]
blahmonkey: Good catch. Fixed in e621752 — moved `_processed_dust_close_ids.add(dust_close_id)` into the same `_dedup_lock` block as the check so reservation is atomic. The two trailing `with self._dedup_lock: ...add(...)` blocks (after missing-position branch and after successful close) are removed since the id is owned upfront. Only one caller now passes the dedup gate; subsequent state mutation is single-writer.

Regression test added — `test_close_as_dust_concurrent_calls_same_id_no_double_credit`: 8 threads gated on a `threading.Event` race into the same `dust_close_id`. Asserts exactly one True result, position deleted exactly once, `account_dust` credited exactly once.

Class sweep: the same non-atomic check-then-act pattern exists on `_processed_outcome_ids` in `process()` at lines 113-148 (in `main` pre-existing). Same class of defect. Out of scope for this PR but worth a follow-up — happy to open a tracking issue if you want it separately.

nexus/infrastructure/praxis_connector/outcome_processor.py
blahmonkey: Documented in e621752. Added an inline comment in the dust branch of `_reduce_position` explaining the capital-aggregate seam: `order_exit` was called earlier with only the sold qty (`avg_cost_basis * fill_size`), the residue's cost basis (`avg_cost_basis * residue`) stays in `position_notional` / `per_strategy_deployed` until the next `reconcile_at_boot`, and the quote-side residue is intentionally left to reconcile because `account_dust` is base-asset only.

Class sweep: same seam exists at `close_as_dust` (sibling thread on line 640). Both sites are documented in this commit — `_reduce_position` via inline comment, `close_as_dust` via expanded docstring. No behavior change; just made the implicit contract explicit.

nexus/infrastructure/praxis_connector/outcome_processor.py
blahmonkey: Documented in e621752. Expanded the `close_as_dust` docstring with the seam: no `order_exit` is called here (no venue order existed), so the full `avg_cost_basis * size` stays in `position_notional` / `per_strategy_deployed` until `reconcile_at_boot` rebuilds from live positions. Sub-lot by construction (B2 only fires when the lot-snapped exit qty is below `LOT_SIZE.minQty`), so the dust-sized cost-basis residue self-corrects at the next reconcile. The implicit caller contract is now explicit in the docstring.

Class sweep: sibling seam in `_reduce_position` documented with an inline comment in the same commit (see the sibling thread on line 527). No behavior change at either site — just making the existing reconcile-dependency explicit so the next reader doesn't have to trace it.

zero-bang (approved):

PR #84, opened 2026-06-10
fix: EXIT ACKs no-op the capital lifecycle instead of failing order_ack (v0.58.0)

blahmonkey:

## What does this PR change?

EXIT ACK outcomes structurally fail `order_ack failed: order '<command_id>' not found` because `CapitalController.order_ack` transitions a `TrackedOrder` from IN_FLIGHT to WORKING — and `TrackedOrder`s are only created by `send_order` from a capital reservation, which EXIT commands never make. The failure predates this release: the historical `OutcomeLoop` queue lag meant EXIT ACKs were rarely processed promptly enough to surface in volume. Once outcome processing kept pace with strategy ticks during the Praxis paper-trade validation on 2026-06-10, 232+ failures surfaced in under an hour.

### `fix:` `_handle_ack` becomes context-aware; EXIT ACKs are a no-op success

[`OutcomeProcessor._handle_ack`](nexus/infrastructure/praxis_connector/outcome_processor.py) now accepts the `OrderContext` and returns success without touching the capital controller when `context.is_entry` is `False` — the venue acknowledged the sell and there is no capital lifecycle to transition. ENTER ACKs are unchanged: they still call `CapitalController.order_ack` and transition IN_FLIGHT to WORKING with `capital_updated=True`.

The no-op success also restores the normal `OutcomeAcked` spine append on the Praxis side (the launcher only appends the ack on `result.success`), so EXIT ACK outcomes no longer accumulate un-acked in the event spine and replay cleanly at boot.

## Validation on prod paper-trade

Deployed to the Praxis paper-trade host (binsim backend) as a mounted-module patch and observed for a full trading cycle (epoch 22, 2026-06-10): **1,284 SELL fills processed with zero `order_ack` failures**, versus 232+ failures in the equivalent window of the prior epoch running without the fix. ENTER ACK capital transitions verified intact in the same window (zero mutation failures across 4,600+ commands).

## Tests

3 new cases in [`tests/test_outcome_processor.py::TestOutcomeProcessorAck`](tests/test_outcome_processor.py):

- `test_exit_ack_noop_success` — EXIT ACK returns `success=True` with `capital_updated=False` and `position_updated=False`
- `test_exit_ack_does_not_touch_capital` — EXIT ACK leaves `in_flight_order_notional` and `working_order_notional` untouched while an unrelated ENTER order is in flight
- `test_entry_ack_still_transitions_capital` — ENTER ACK still moves the full reserved total from in-flight to working (guarded by an explicit `in_flight_before > _ZERO` precondition so the assertion cannot pass trivially)

`ruff` clean, `mypy` clean (75 source files), **1,491 tests pass** / 17 skipped (was 1,488 / 17).

Bumps version 0.57.0 → 0.58.0. Companion Praxis-side changes (synchronous accounting at the outcome route boundary + dust-close gate) ship separately and will pin this release.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -714,3 +714,17 @@
 ### Fix
 
 - Closes [Vaquum/Nexus#82](https://github.com/Vaquum/Nexus/issues/82). Companion Praxis-side fix at [Vaquum/Praxis#142](https://github.com/Vaquum/Praxis/issues/142) — the launcher changes that compute `intended_full_close`, carry it through `_build_order_context`, and call `OutcomeProcessor.close_as_dust` in the rejection branch ship there. Nexus alone is dormant until that companion lands (the new field defaults to `False` so behavior is unchanged on every existing caller).
+
+## v0.58.0 on 10th of June, 2026
+
+### NOTE
+
+- During the Praxis v0.76.0 paper-trade validation (epoch 22, 2026-06-10), once outcome processing kept pace with strategy ticks, every EXIT ACK surfaced as `order_ack failed: order '<command_id>' not found` — 232+ failures in under an hour. Root cause: `CapitalController.order_ack` transitions a `TrackedOrder` from IN_FLIGHT to WORKING, but `TrackedOrder`s are only created by `send_order` from a capital reservation, and EXIT commands never reserve capital. The failure is structural — an EXIT ACK can never find an order record. It predates this release; the historical `OutcomeLoop` queue lag simply meant EXIT ACKs were rarely processed promptly enough for the failure to surface in volume.
+
+### Update
+
+- Update [`OutcomeProcessor._handle_ack`](nexus/infrastructure/praxis_connector/outcome_processor.py) to accept the `OrderContext` and return a no-op success for `is_entry=False` outcomes — the venue acknowledged the sell and there is no capital lifecycle to transition. ENTER ACKs are unchanged: they still call `CapitalController.order_ack` and transition IN_FLIGHT to WORKING. The no-op success also restores the normal `OutcomeAcked` spine append on the Praxis side, so EXIT ACK outcomes no longer accumulate un-acked in the event spine and replay cleanly at boot
+
+### Add
+
+- Add 3 tests to [`tests/test_outcome_processor.py::TestOutcomeProcessorAck`](tests/test_outcome_processor.py): EXIT ACK returns success with `capital_updated=False`; EXIT ACK leaves `in_flight_order_notional` and `working_order_notional` untouched while an unrelated ENTER order is in flight; ENTER ACK still transitions the full reserved total from in-flight to working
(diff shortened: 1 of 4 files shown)


zero-bang (dismissed):
**Approve.** Correct, minimal fix with a verified root cause.

## Root cause — verified
I traced the claim independently. EXIT actions bypass the CAPITAL validation stage (`nexus/core/validator/pipeline_executor.py:150-165`), so `check_and_reserve` is never called for an EXIT → no reservation → `bridge_to_capital`/`send_order` never creates a `TrackedOrder` → `CapitalController.order_ack(command_id)` could never find a record for an EXIT command. The pre-fix `order_ack failed: order '<id>' not found` was therefore structural, exactly as described. Routing EXIT ACKs to a no-op success is the right fix.

## Correctness
- Dedup interaction is right: the no-op returns `success=True`, so the `outcome_id` is added to `_processed_outcome_ids` (replay-safe) while `capital_updated`/`position_updated` stay `False`, so the launcher won't fire `append_mutation`. This is also what restores the `OutcomeAcked` spine append on the Praxis side (launcher appends only on `result.success`).
- `is_entry` is the same trust boundary the rest of the processor already routes on (`_handle_fill`, `_compute_exit_cost_basis`), so no new assumption is introduced.
- Signature change is on a private method with a single internal caller (`process`); no other callers to update.
- Tests are meaningful — `test_entry_ack_still_transitions_capital` guards against a trivially-passing assertion with the explicit `in_flight_before > _ZERO` precondition.

I ran the full `tests/test_outcome_processor.py` on the PR branch locally: **67 passed** (incl. the 3 new cases). ruff/CHANGELOG/version bump all in order.

## CI disposition (required-check ledger)
Ruff passes. Tests, Mypy, and CodeQL all fail **identically** in the `Install dependencies` step: `fatal: couldn't find remote ref refs/tags/v4.0.1` for `vaquum_limen @ git+https://github.com/Vaquum/Limen@v4.0.1`. That Limen pin is **not touched by this PR**, and `main` is green with the same pin — so this is a transient dependency-fetch failure, not the diff. **It is safe to ignore for the diff's merits, but a green gate has never actually validated this change** — please re-run CI and confirm Tests/Mypy/CodeQL pass before merging (the verification value of those gates is currently zero).

## Engineering note
The prod-validation claim (1,284 SELL fills, zero failures) has no automated regression gate — `test_exit_ack_noop_success` locks the unit contract, but nothing asserts at the launcher/spine seam that an EXIT ACK actually produces an `OutcomeAcked` append. Worth a carry-forward integration test on the Praxis side when the companion lands.

## Forward-looking question
Next predictable failure mode: a **partial** EXIT FILL still routes through `_handle_fill` → `order_exit` (capital path), which is correct — but is there any outcome type for an EXIT that *does* expect a tracked-order transition (e.g. an EXIT CANCEL/REJECT arriving for a venue order that was somehow reserved through a non-standard path)? `_handle_reject`/`_handle_cancel` already tolerate `not context.is_exit` only for the failure branch; confirm no future EXIT-with-reservation path (e.g. margin/borrow flows) reintroduces a real `TrackedOrder` that this no-op would then silently skip.

zero-bang (approved):

Issue #88, opened 2026-06-12
serialize_state reads InstanceState without locks — torn snapshots possible


blahmonkey:
## Problem

\`wal_codec.serialize_state\` iterates \`state.positions\`, \`state.account_dust\`, and reads capital/risk/mode fields with no locking. Mutators run on other threads (the OutcomeLoop worker, the Praxis trading thread's synchronous accounting, the submitter's \`pending_exit\` writes). Consequences:

- \`RuntimeError: dictionary changed size during iteration\` on the dict snapshots
- torn cross-field reads of \`CapitalState\` (fields are mutated under \`CapitalController._lock\`, which the serializer does not take) — a serialized record can violate decode invariants and poison recovery up to that point

Current mitigations are call-site discipline: the Praxis launcher wraps its post-outcome \`append_mutation\` in \`positions_lock\` (Praxis v0.77.0) and \`SnapshotScheduler\` takes \`positions_lock\` + a capital lock context. The codec itself remains unsafe for any future caller.

## Fix direction

Move the locking into the codec/StateStore layer so every caller is safe by construction — e.g. \`StateStore\` methods acquire the relevant locks (injected at construction) around the serialize, or \`serialize_state\` gains an explicit lock-bundle parameter. Call-site wrapping then becomes redundant and can be removed.

Issue #87, opened 2026-06-12
Submit timeout is classified SUBMIT_FAILED while the command may still execute


blahmonkey:
## Problem

In \`nexus/strategy/action_submit.py\`, any exception from \`praxis_outbound.send_command\` — including \`TimeoutError\` — takes the SUBMIT_FAILED branch: \`_release_granted_reservation\` runs and no registry entry is created. But a timeout is not proof of failure: \`run_coroutine_threadsafe(...).cancel()\` can only interrupt Praxis's \`submit_command\` coroutine at its single await point (the \`CommandAccepted\` spine append); once past it, the queue insertion and registration are synchronous and the command **will execute**. The fill then arrives for a command Nexus released capital for and forgot.

Suspected fallout already observed: an epoch-20 (2026-06-10) WAL record with negative \`working_order_notional\` that failed CapitalState's decode invariant — consistent with a reservation released by the timeout path and decremented again by the late fill's accounting.

## Fix direction (requires the deterministic-command-id issue)

- A timeout (or ambiguous transport error) after handoff becomes \`SUBMISSION_UNKNOWN\`, not SUBMIT_FAILED: the reservation is kept, the registry entries (created pre-handoff under the companion issue) are retained, and reconciliation resolves the command by its deterministic id — either outcomes arrive (normal processing) or a probe/timeout sweep declares it dead and releases.
- Only failures that provably occurred before handoff (local validation, translate errors) release the reservation immediately.

Companion: Vaquum/Nexus#86.

blahmonkey:
Implemented; closing as done.

`SubmissionStatus.SUBMISSION_UNKNOWN` exists (`nexus/strategy/action_submit.py:76`) and the timeout branch does what the fix direction asks: it calls `registration.mark_unknown(e)`, **retains** the granted reservation and the pre-handoff registry entries, and returns `SUBMISSION_UNKNOWN` carrying the deterministic `command_id` so reconciliation can resolve the command either way. The reservation is released only for failures that provably occurred before handoff — context-build, validation, translate, and `pre_register` itself.

Landed in `c010d7e`, alongside the deterministic-id work this depended on.

The semantics are gated on `PraxisOutbound.supports_command_id`; when false the legacy submit-then-register path still applies, since pre-registering against an id the receiver may not honour would be unsafe. Verified live against the installed pair: `_command_id_support(ExecutionManager.submit_command)` returns `(True, True)`, so the `SUBMISSION_UNKNOWN` path is active in production.

[...]

Issue #86, opened 2026-06-12
Command identity is born inside Praxis after the point of no return


blahmonkey:
## Problem

\`PraxisOutbound.send_command\` does not transmit \`TradeCommand.command_id\`. Praxis's \`ExecutionManager.submit_command\` mints its own \`command_id = str(uuid.uuid4())\` and the only way the authoritative id reaches Nexus is the coroutine's return value. Two structural consequences:

1. **Registration must happen after \`send_command\` returns.** The launcher's submitter can only register \`command_strategy_ids\` / \`command_contexts\` / the capital \`TrackedOrder\` once it has the Praxis-minted id — so a fast venue can deliver translated outcomes before any registry knows the command. This registration-gap race produced the lost-fill incidents of 2026-06-11 (epoch 24: two venue-filled EXITs never accounted). v0.59.0's unresolved-outcome retry and Praxis v0.78.0's skip telemetry treat the symptom; under heavy EXIT cycling the race fires ~170×/cycle and heals, but the design remains backwards.

2. **A submit timeout loses the identity forever.** \`send_command\` raising \`TimeoutError\` after the Praxis coroutine has queued the command (the post-await section of \`submit_command\` is synchronous and cannot be interrupted by \`future.cancel()\`) leaves a command that WILL execute with no Nexus-side identity, no registry entry, and a released reservation — see the companion timeout issue.

## Fix direction

- Nexus owns the authoritative command id: \`PraxisOutbound.send_command\` transmits \`command.command_id\`.
- \`ExecutionManager.submit_command\` accepts an optional caller-supplied \`command_id\` (validated against accepted/active registries for duplicates) and uses it instead of minting.
- The launcher registers strategy mapping, \`OrderContext\`, and capital tracking **before** outbound handoff; only failures that provably occurred before handoff may roll registration back.

Companion: the timeout-as-unknown issue (same release), and the Praxis-side acceptance change.

blahmonkey:
Implemented; closing as done.

All three fix-direction items are in `main`:

- **Nexus transmits the authoritative id** — `PraxisOutbound.send_command` sets `submit_kwargs['command_id'] = command.command_id` (`nexus/infrastructure/praxis_connector/praxis_outbound.py:177`), and a `CommandIdMismatchError` guard rejects a receiver that returns a different id than it was given.
[...]

PR #90, opened 2026-06-12
fix: poll for IPC refcount drain in the broken-pool resilience test

### LINKED ISSUE #89: test_broken_pool_submit_does_not_kill_scheduler is timing-flaky in CI
## Problem

\`tests/test_predict_loop.py::TestPredictLoopResilience::test_broken_pool_submit_does_not_kill_scheduler\` waits a fixed \`time.sleep(0.1)\` before asserting \`all(count == 0 for count in loop._ipc_refs.values())\`. On loaded CI runners the refcount release can land after the fixed sleep, failing the run (observed 2026-06-10; passed on rerun and passes locally 3/3). 

## Fix direction

Replace the fixed sleep with a bounded poll: re-check the refcounts until they drain or a generous deadline (~2s) expires, asserting on the final state. Same pattern as the OutcomeLoop retry tests.

blahmonkey:

## What does this PR change?

Closes #89. `test_broken_pool_submit_does_not_kill_scheduler` asserted the `_ipc_refs` rollback immediately after observing the submit attempt; on loaded CI runners the scheduler thread's rollback can land after the fixed wait, failing the run (observed 2026-06-10, passed on rerun, passes locally consistently). The assertion now polls for the refcount drain with a 2s deadline before asserting the final state — the same pattern as the `OutcomeLoop` retry tests.

Test-only change: no CHANGELOG entry / version bump per house convention for non-code changes.

`ruff` clean; 14 predict-loop tests pass.

## Checklist

- [ ] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/dev-docs/blob/main/src/Writing-Docstrings.md) (for any changed public functions/classes)
- [ ] I updated CHANGELOG.md (unless only docs or other non-code aspect was changed)
- [ ] I updated pyproject.toml (unless only docs or other non-code aspect was changed)
- [x] I linked issue to auto-close on merge (e.g., “Fixes #123”) when applicable

### DIFF
--- a/tests/test_predict_loop.py
+++ b/tests/test_predict_loop.py
@@ -531,6 +531,15 @@ class TestPredictLoopResilience:
             assert loop._scheduler_thread is not None
             assert loop._scheduler_thread.is_alive()
             assert runner.dispatch_signal.called is False
+
+            deadline = time.monotonic() + 2.0
+
+            while (
+                any(count != 0 for count in loop._ipc_refs.values())
+                and time.monotonic() < deadline
+            ):
+                time.sleep(0.02)
+
             assert all(count == 0 for count in loop._ipc_refs.values())
 
             loop.stop()


zero-bang (approved):
LGTM — clean, minimal flakiness fix.

The poll-then-assert replaces an immediate refcount assertion that raced the scheduler thread's rollback. Verified it's structurally identical to the established `OutcomeLoop` retry pattern (`tests/test_outcome_loop.py:210-216`): same 2.0s deadline, 0.02s sleep, two-condition `while`. The unlocked read of `_ipc_refs.values()` is safe here because the single wired sensor's key set is fixed (only int counts mutate), so no size-change-during-iteration risk — same profile as the original assertion, no regression.

Edge case: empty `_ipc_refs` still passes vacuously, identical to prior behavior. Test-only change, so skipping CHANGELOG/version bump per house convention is correct.

Engineering note (non-blocking): the 2.0s deadline is now duplicated across many of these timing tests as a magic literal. If CI runners get slower, a single shared constant would be one place to bump rather than hunting each call site. Worth keeping in mind, not worth doing here.

Forward-looking: the one behavioral tradeoff is that a genuine regression in the rollback logic (refs never draining) now costs a full 2s before this test fails, instead of failing instantly — acceptable for the flakiness it removes.

Note: Tests and CodeQL checks were still pending at review time; Ruff and Mypy pass. The diff is correct on inspection — deferring final confirmation to the Tests run.

blahmonkey:
Folding into the combined Nexus train (identity #86 + timeout-as-unknown #87 + codec locking #88 + this test fix #89) — one PR per repo per batch. The commit carries over unchanged.

PR #98, opened 2026-06-16
fix: preserve Nexus positions on reconcile divergence; key on pos.trade_id (v0.64.0)

blahmonkey:

## What does this PR change?

Fixes the Nexus half of a restart durability bug where a process restart dropped open positions: boot reconciliation silently deleted any Nexus position the Praxis snapshot lacked, orphaning the venue holding.

### Preserve, don't evict

`_reconcile_capital` previously evicted Nexus positions absent from the Praxis snapshot ("Praxis truth wins") and checkpointed the deletion. On a restart that discarded a live position while the venue still held it. It now **preserves** the position and **fails closed** (raises `StartupError`) so the divergence is investigated rather than resolved by data loss. Pairs with the Praxis change ([Vaquum/Praxis#155](https://github.com/Vaquum/Praxis/pull/155)) that stops entry fills from emitting `TradeClosed`, which is what made Praxis under-report open positions on replay.

### Key on `pos.trade_id`

The reconcile parse loop matched Praxis positions by unpacking the snapshot tuple key and reading element `[1]`. Praxis keys positions `(trade_id, account_id)`, so it read the **account_id** as the trade_id and every Praxis position mis-bucketed and never matched the Nexus side. It now matches on `pos.trade_id` directly, independent of key ordering.

TD-095 records a follow-up: distinguish a venue-recorded close (apply it) from a lost position (fail closed), so a restart racing a legitimate close can self-recover.

## Tests

Rewrote the two eviction tests to the fail-closed contract (raises `StartupError`, position preserved) and added a keying regression test that feeds the real `(trade_id, account_id)` snapshot order. `ruff` + `mypy` clean, **1,447 tests pass**. Bumps 0.63.1 → 0.64.0.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -807,3 +807,10 @@
 ### Update
 
 - Prune the technical-debt register in [`docs/TechnicalDebt.md`](docs/TechnicalDebt.md): collapse code-verified-resolved (TD-004, TD-005, TD-006, TD-010, TD-011, TD-013, TD-015, TD-016, TD-017, TD-018, TD-023) and Conduit-obsoleted (TD-019, TD-020, TD-021, TD-022, TD-087, TD-088, TD-089) entries to heading-only status markers, and mark TD-051 RESOLVED (`_reduce_position` already nets exit fees at [`outcome_processor.py`](nexus/infrastructure/praxis_connector/outcome_processor.py) via `realized_pnl = gross_pnl - actual_fees`)
+
+## v0.64.0 on 16th of June, 2026
+
+### Fix
+
+- Stop boot reconciliation from silently deleting a Nexus position absent from the Praxis snapshot. `_reconcile_capital` previously evicted such positions ("Praxis truth wins") and checkpointed the deletion, so a restart could discard a live position and orphan the venue holding. It now preserves the position and fails closed (raises [`StartupError`](nexus/startup/error.py)) so the divergence is investigated rather than resolved by data loss. Pairs with the Praxis fix that stops entry fills from emitting `TradeClosed` (which made Praxis under-report open positions on replay)
+- Match Praxis positions on `pos.trade_id` rather than unpacking the snapshot tuple key in `_reconcile_capital`. Praxis keys positions `(trade_id, account_id)` but reconcile read element `[1]` (the account_id) as the trade_id, so every Praxis position mis-bucketed and never matched the Nexus side
--- a/docs/TechnicalDebt.md
+++ b/docs/TechnicalDebt.md
@@ -996,3 +996,16 @@ Inside the terminal-release path, `order_fill` calls `_adjust_strategy_deployed(
 **When to fix**: Never on its own. Roll into TD-093 if/when the branch is being touched anyway; otherwise the two-call shape mirrors the natural read order (apply fee delta, then apply residual release) and is easier to reason about.
 
 **Migration**: Combine the two `_adjust_strategy_deployed` calls in the terminal-release branch into one with the summed delta. Strictly cosmetic.
+
+---
+
+## TD-095: Boot reconcile fails closed even on a position the venue legitimately closed
+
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -807,3 +807,10 @@
...
+
+## v0.64.0 on 16th of June, 2026
+
+### Fix
+
+- Stop boot reconciliation from silently deleting a Nexus position absent from the Praxis snapshot. `_reconcile_capital` previously evicted such positions ("Praxis truth wins") and checkpointed the deletion, so a restart could discard a live position and orphan the venue holding. It now preserves the position and fails closed (raises [`StartupError`](nexus/startup/error.py)) so the divergence is investigated rather than resolved by data loss. Pairs with the Praxis fix that stops entry fills from emitting `TradeClosed` (which made Praxis under-report open positions on replay)
+- Match Praxis positions on `pos.trade_id` rather than unpacking the snapshot tuple key in `_reconcile_capital`. Praxis keys positions `(trade_id, account_id)` but reconcile read element `[1]` (the account_id) as the trade_id, so every Praxis position mis-bucketed and never matched the Nexus side
```
zero-bang, line 816: Non-blocking, but worth fixing in this PR since you're establishing the key order here: this entry (correctly) states Praxis keys positions `(trade_id, account_id)`, yet the consumed function's own docstring still claims the opposite — `nexus/infrastructure/praxis_connector/praxis_outbound.py:364` reads `Positions keyed by (account_id, trade_id) tuples`. That stale docstring is precisely the trap that produced this bug: the next engineer who trusts it and writes `account_id, trade_id = key` reintroduces the mis-bucketing you just fixed. Your new code sidesteps ordering entirely by keying on `pos.trade_id` (good), but the contract doc on the seam still points the wrong way. Suggest correcting `praxis_outbound.py:364` to `(trade_id, account_id)` (and the `dict[tuple[str, str], Any]` is order-agnostic so no type change needed).

zero-bang (commented):
## Review — solid durability fix

The core change is correct and the reasoning is sound. Two real bugs are addressed cleanly:

1. **Keying on `pos.trade_id`** is the right move — it's independent of tuple ordering, so it can't silently mis-bucket again. The old `_, trade_id = key` read element `[1]`; since every position for one account shares the same `account_id`, the old `praxis_by_trade_id` dict collapsed *all* Praxis positions into a single entry, which meant reconcile was evicting nearly every Nexus position. This was worse than the description implies, and the fix resolves it at the root.
2. **Preserve-and-fail-closed** is the correct default for an orphan/data-loss risk: halting for investigation beats silently dropping a live venue holding. TD-095 honestly records the accepted tradeoff (a restart racing a legitimate close now halts instead of self-recovering).

The rewritten tests match the new contract, and the new keying-regression test genuinely exercises the real `(trade_id, account_id)` order — it would fail under the old code (position evicted), so it's a meaningful guard, not a tautology.

One non-blocking follow-up below (stale docstring on the consumed function). Not requesting changes.

**Checks:** Ruff and Mypy pass; CodeQL and Tests were still pending at review time — none red. Worth confirming Tests goes green before merge.

**Engineering note:** the fail-closed choice trades availability for safety. That's the right call here, but it does shift a recoverable case into an operator-paging halt — see the forward-looking note below.

**Forward-looking:** the next predictable failure mode is a TD-095 trip during the soak — an unattended restart that races a *legitimate* venue close (terminal outcome not yet processed pre-restart) will now `raise StartupError` and refuse to boot, paging an operator for what is actually a benign, already-closed position. If that proves more than rare, the TD-095 migration (Praxis exposing whether a `trade_id` has a recorded terminal close) becomes load-bearing rather than nice-to-have.

blahmonkey:
Re: the stale key-order docstring — fixed in 6b0bd00: `pull_positions` docstring corrected to `(trade_id, account_id)` with a note that consumers should match on `pos.trade_id`, not tuple position. Thread resolved.

zero-bang (approved):

PR #99, opened 2026-06-16
fix: snap sub-ULP capital residues so restart recovery cannot brick (v0.65.0)

blahmonkey:

## What does this PR change?

Fixes a restart-durability bug found while validating the position-recovery fix: after live fee-bearing trading, restarting the process crash-looped the Nexus instance during state recovery.

### Root cause

`CapitalController.order_fill` releases a terminal fill's working aggregate via the proportional-fee Decimal division in `TrackedOrder.remaining_total`, which can leave `working_order_notional` at a sub-ULP negative residue (observed: `-1E-27`), with a mirror `fee_reserve = +1E-27`. `CapitalState.__post_init__` rejects any negative aggregate, so `deserialize_state` raised `Malformed WAL codec payload` → `StartupError` in `_recover_state` → the instance crash-looped on boot. The residue was latent until the Praxis fee fix made `actual_fees` non-zero (at `fee_rate=0` the reserve/release netted to exactly zero).

### Fix

- **Mutation site** (`order_fill`): snap `working_order_notional`, `fee_reserve`, and `per_strategy_deployed` residues within `SUB_ULP_TOLERANCE` to zero. The `fee_reserve` snap previously handled only tiny negatives; `_adjust_strategy_deployed` now treats sub-tolerance magnitudes as zero rather than logging a false underflow (genuine underflows beyond tolerance still warn).
- **Recovery shim** (`wal_codec._decode_capital_state`): snap sub-tolerance *negative* persisted aggregates to zero before constructing `CapitalState`, so snapshots already carrying a residue boot. The domain invariant stays strict — only sub-`SUB_ULP_TOLERANCE` negatives are snapped; meaningful negatives still raise.
- `SUB_ULP_TOLERANCE` is now a single source in `capital_state` (was forked into the codec).

TD-096 records a follow-up: collapse the two-step terminal residual release in `order_fill` into one exact decrement. The shipped position-recovery fix (Praxis 0.83.0 / Nexus 0.64.0) is independent and unaffected — this bug is upstream of reconcile.

## Tests

Added codec tests asserting a sub-ULP negative residue decodes to zero and a meaningful negative still raises. `ruff` + `mypy` clean, **1,449 tests pass**. Bumps 0.64.0 → 0.65.0.

## Checklist

- [] I updated `/docs` (if behavior/API/config/user/etc changed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -814,3 +814,15 @@
 
 - Stop boot reconciliation from silently deleting a Nexus position absent from the Praxis snapshot. `_reconcile_capital` previously evicted such positions ("Praxis truth wins") and checkpointed the deletion, so a restart could discard a live position and orphan the venue holding. It now preserves the position and fails closed (raises [`StartupError`](nexus/startup/error.py)) so the divergence is investigated rather than resolved by data loss. Pairs with the Praxis fix that stops entry fills from emitting `TradeClosed` (which made Praxis under-report open positions on replay)
 - Match Praxis positions on `pos.trade_id` rather than unpacking the snapshot tuple key in `_reconcile_capital`. Praxis keys positions `(trade_id, account_id)` but reconcile read element `[1]` (the account_id) as the trade_id, so every Praxis position mis-bucketed and never matched the Nexus side
+
+## v0.65.0 on 16th of June, 2026
+
+### Add
+
+- Add `test_decode_capital_state_snaps_subulp_negative_residue` and `test_decode_capital_state_rejects_meaningful_negative` to [`tests/test_wal_codec.py`](tests/test_wal_codec.py)
+- Add TD-096 recording the structural follow-up (collapse the two-step terminal residual release in `order_fill` into one exact decrement)
+
+### Fix
+
+- Snap sub-ULP Decimal residues in `CapitalController` capital aggregates to zero so a restart cannot brick on recovery. After a fee-bearing terminal fill, the reserve/release arithmetic (proportional-fee division in `TrackedOrder.remaining_total`) could leave `working_order_notional` at a sub-ULP negative (e.g. `-1E-27`); `CapitalState.__post_init__` rejects any negative, so `deserialize_state` failed and the instance crash-looped on boot. `order_fill` now snaps `working_order_notional`, `fee_reserve`, and `per_strategy_deployed` residues within `_SUB_ULP_TOLERANCE` to zero (the `fee_reserve` snap previously handled only tiny negatives; `_adjust_strategy_deployed` now treats sub-tolerance magnitudes as zero rather than logging a false underflow). Latent until the Praxis fee fix made `actual_fees` non-zero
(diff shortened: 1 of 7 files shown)


zero-bang (dismissed):
LGTM — approving.

The diagnosis and fix are precise. Verified the reasoning end-to-end:

- **Live snaps are aggregate-scoped.** Both `abs(working_order_notional) <= _SUB_ULP_TOLERANCE` and the `fee_reserve` snap fire only when the *total* field is within sub-ULP of zero, so they can't mask a genuine balance — a residue left while other orders are still working stays embedded in a large positive total (which boots fine) and only gets snapped once the aggregate returns near zero. The mirrored `working_order_notional = -1E-27` / `fee_reserve = +1E-27` residues net to zero in `available`, so snapping both to zero leaves the derived value unchanged.
- **`fee_reserve` snap broadening is correct.** Moving from negative-only to `abs() <=` is right because the mirror residue can be positive; sub-ULP positive is below any meaningful money.
- **Codec shim is correctly asymmetric.** `-_SUB_ULP_TOLERANCE <= value < _ZERO` clamps negatives only (positives already satisfy `__post_init__`) and leaves meaningful negatives intact so genuinely-broken state still fails closed. Both new tests confirm the boundary, and `-0.5` correctly raises.
- **`_adjust_strategy_deployed`** now folds the `== _ZERO` pop into the `abs(updated) <= tol` branch and suppresses the false underflow warning for sub-tolerance magnitudes while preserving the real-underflow warning. Correct.

CI disposition: Mypy and Ruff pass; Tests and CodeQL were pending at review time — I ran the touched suites locally (`tests/test_wal_codec.py` 43 passed, `tests/test_capital_controller.py` 129 passed), so the pending Tests check is not a concern.

Two non-blocking notes:

1. The `per_strategy_deployed` codec clamp can leave a `Decimal(0)` entry where the live `_adjust_strategy_deployed` would `pop` it. Harmless (a zero entry is arithmetically and semantically identical to absent), and it only arises for pre-fix snapshots — flagging only so it's a conscious choice. If you want full parity with the live invariant you could drop clamped-to-zero per-strategy entries, but it's not worth churning the diff.

2. Forward-looking failure mode: the snap is an *absolute* `1E-12` threshold (already noted in the constant's comment and a TD). The next predictable break is a deployment trading sub-cent notionals (or notional > 10^15) where `1E-12` is no longer "sub-ULP" relative to a real increment — at that point a genuine residual could exceed the tolerance and re-introduce the brick, or a meaningful balance could fall under it. The TD-096 follow-up (one exact decrement) removes the residue at the source and is the durable fix; worth prioritizing if/when sub-cent venues come online.

Nice catch on this being latent until the Praxis fee fix made `actual_fees` non-zero.

blahmonkey:
Thanks for the thorough review. On the two notes:

1. **per_strategy_deployed Decimal(0) on decode** — I tried dropping clamped-to-zero entries for parity with the live `_adjust_strategy_deployed` pop, but an existing round-trip test (`test_serialize_round_trips_capital_per_strategy_deployed`) intentionally pins codec fidelity for an exact-zero entry, so dropping broke it. Reverted (1b30e96) to keep faithful round-trip — the leftover `Decimal(0)` is harmless (identical to absent) as you noted, and codec fidelity is the better invariant to preserve.

2. **Absolute 1E-12 threshold** — agreed; tracked as TD-096 (collapse the two-step terminal residual release into one exact decrement, removing the residue at source). That is the durable fix and should land before any sub-cent-notional venue. The current constant comment + TD note the relative-threshold limitation.

Both non-blocking; no functional change to the fix itself.

zero-bang (approved):

PR #100, opened 2026-06-19
fix: make outcome dedup durable across restart (v0.66.0)

blahmonkey:

## What does this PR change?

Makes the outcome-dedup durable so a process restart cannot double-apply an already-applied outcome. Addresses the durable applied-outcome marker required by Nexus#86 / TD-086.

### Root cause

`OutcomeProcessor` held the processed `outcome_id`s in a process-local set, lost on restart. The mutation those ids guard is persisted by the launcher's `append_mutation`. So if the process was killed between `append_mutation` (mutation durable) and the Praxis `OutcomeAcked` (delivery confirmed), a boot replay of that un-acked-but-already-applied outcome re-entered `_handle_fill` and mutated capital and position a **second** time — a durable mutation paired with a restart-volatile dedup set.

### Fix

- The dedup sets now live on `InstanceState` (`processed_outcome_ids`, `processed_dust_close_ids`) instead of `OutcomeProcessor`-local sets, so the successful id is persisted **atomically with the mutation it guards** (the launcher's `append_mutation` writes the whole state) and reconstructed by `StateStore.recover`. A boot replay of an already-applied-and-persisted outcome is recognised in `process()` and returns a no-op success.
- Serialized via the existing `wal_codec` v1 payload (sorted lists) with `.get` back-compat, so pre-upgrade snapshots load unchanged.
- `close_as_dust` gets the same durability.

This satisfies the acceptance-addendum requirement that `recover()` rebuild the applied-outcome marker, so the Praxis boot-replay (TD-052) can land safely against this dedup. The in-process mid-`_handle_fill` partial-mutation guard remains tracked in TD-086; the new unbounded-growth follow-up is recorded as TD-097.

## Tests

Five new tests: durable round-trip + default-empty + pre-#86 back-compat (`test_wal_codec`); cross-restart no-double-apply + dust-dedup-survives-restart (`test_outcome_processor`). `ruff` + `mypy` clean, **1,454 tests pass**. Bumps 0.65.0 → 0.66.0.

## Checklist

- [x] I updated `/docs` (if behavior/API/config/user/etc changed)


nexus/infrastructure/praxis_connector/outcome_processor.py
```diff
@@ -145,7 +153,7 @@ def process(
 
         if result.success:
             with self._dedup_lock:
-                self._processed_outcome_ids.add(outcome.outcome_id)
+                self._state.processed_outcome_ids.add(outcome.outcome_id)
```
zero-bang, line 163: The dedup-set add lives under `_dedup_lock` only, but `serialize_state` reads `state.processed_outcome_ids` under the snapshot locks (`positions_lock` + `capital_lock`, via `StateStore._state_read_guard`) — a different lock domain. Note `account_dust`, also persisted state mutated by this same class, is deliberately written *inside* `self._positions_cm` (see `_reduce_position` ~L560 and `close_as_dust` ~L678) precisely so a concurrent `checkpoint()` / `append_mutation()` can't serialize it inconsistently. The new dedup fields break that established pattern.

Concrete consequence: `order_fill` commits the capital mutation and releases `CapitalController._lock` *before* control returns here for the add. A `SnapshotScheduler` `checkpoint()` firing in that gap holds `positions_lock`+`capital_lock`, so it can serialize the mutated capital while `outcome_id` is not yet in the set — and `checkpoint()` truncates the WAL. A crash between that truncate and the paired `append_mutation` then recovers a snapshot with capital mutated but no dedup marker → replay re-applies. That is the exact double-apply this PR sets out to close, just narrowed to a checkpoint-interleave window rather than eliminated.

The `set(state.processed_outcome_ids)` copy itself is atomic under the GIL (C-level `set_merge` holds the GIL), so this is a consistency/atomicity-window concern, not a 'Set changed size during iteration' crash — worth verifying if a free-threaded build is ever on the table.

Fix options: (a) move this `.add` (and the `close_as_dust` one at L675) under `self._positions_cm` so it shares the lock domain `serialize_state` reads under — mind the `_dedup_lock` vs `positions_lock` ordering since the check+add must stay atomic for the PR #83 concurrent-same-id guarantee; or (b) if the launcher already holds `positions_lock` across the whole `process()`→`append_mutation` sequence, the window is already closed — in which case a one-line note here stating that invariant would settle it. Which is it?

zero-bang (commented):
## Review summary

Solid, well-scoped change that does what it claims for the common case: moving the dedup sets onto `InstanceState` so the successful id is written by the launcher's `append_mutation(state)` in the same payload as the capital/position mutation, and rebuilt by `StateStore.recover` via `deserialize_state`. I traced the integration points:

- `recover()` (state_store.py:267) replays `STATE_MUTATION` payloads through `deserialize_state`, so the new fields are reconstructed — verified.
- `wal_codec` uses `.get(..., [])` for both keys, so pre-#86 snapshots decode to empty sets — verified, and the back-compat test exercises it directly.
- The only field-by-field `InstanceState(...)` construction is in `wal_codec` (now updated); nothing else rebuilds the dataclass and would silently drop the fields — verified.
- The append_mutation path is internally consistent: every `result.success` adds the id, and every success that mutates capital/position raises `capital_updated`/`position_updated`, which is what gates the launcher's persist. The success-but-no-persist cases (non-entry ACK) are genuine no-ops on replay, so their volatile dedup marker is harmless.

The honest TD-086/TD-097 carry-forward write-ups are appreciated and accurate.

One concern worth the author's attention before merge (inline) — it concerns the lock domain of the dedup-set mutation versus the snapshot-serialization lock, which narrows the exact atomicity guarantee this PR is built on. Submitting as a comment rather than a block because reachability depends on launcher locking I can't see from this repo.

**Engineering note (1.17):** this PR trades a restart-volatile-but-cheap dedup set for a durable-but-unbounded one (TD-097). The migration note (evict outside the Praxis replay horizon) is the right shape, but it depends on a replay-window definition Nexus does not yet own — until Praxis supplies a high-water ack mark, any eviction Nexus does unilaterally risks dropping an id that is still replayable, re-opening the very double-apply this PR closes. Worth making that dependency explicit in TD-097 so the follow-up isn't attempted prematurely.

**Forward-looking question (1.18):** the next predictable failure mode is boot deserialize latency, not capital correctness — a multi-month paper-soak makes `processed_outcome_ids` a five/six-figure sorted list packed into *every* snapshot and *every* STATE_MUTATION WAL entry. The first symptom won't be a wrong number; it'll be `recover()` and per-fill `append_mutation` getting steadily slower as the sorted-list encode grows. Is there a soak-time gate that would catch snapshot size crossing, say, a few MB before it shows up as boot time?

## Open-threads ledger
- Other reviewer threads: none open.
- CI: Mypy pass, Ruff pass, Tests pass. CodeQL still pending at review time — not a red check; no disposition needed beyond noting it should go green before merge.

nexus/infrastructure/praxis_connector/outcome_processor.py
```diff
@@ -656,15 +664,15 @@ def close_as_dust(
...
+            if dust_close_id in self._state.processed_dust_close_ids:
                 _log.debug(
                     'duplicate dust close dropped: dust_close_id=%s trade_id=%s',
                     dust_close_id,
                     trade_id,
                 )
                 return False
 
-            self._processed_dust_close_ids.add(dust_close_id)
+            self._state.processed_dust_close_ids.add(dust_close_id)
```
zero-bang: Same lock-domain point as the `process()` add (see comment on L156): this `processed_dust_close_ids.add` is under `_dedup_lock` but sits *outside* the `self._positions_cm` block that the sibling `account_dust` write (L678+) is correctly wrapped in. `serialize_state` reads this set under `positions_lock`, so the dust dedup marker and its paired `account_dust` mutation are not guaranteed to be serialized together by a concurrent `checkpoint()`. Resolve together with L156.

nexus/infrastructure/praxis_connector/outcome_processor.py
blahmonkey: Addressed in 4cac5a5. Moved both durable dedup adds into the `positions_lock` domain, matching the `account_dust` precedent you cited: `process()` adds under `with self._positions_cm, self._dedup_lock`, and `close_as_dust` folds its check+add into the existing `_positions_cm` block. The `_dedup_lock` is kept *nested* (not replaced) because the PR #83 concurrent-same-id test constructs the processor with no `positions_lock` (`_positions_cm = nullcontext`) and relies on `_dedup_lock` for the check+add atomicity — nesting preserves both. Lock order is positions→dedup everywhere; `_dedup_lock` is acquired nowhere else, so no inversion. New test `test_dust_close_dedup_write_is_gated_by_positions_lock` proves the write blocks while `positions_lock` is held.

On your either/or: it is (a), not (b). The launcher does NOT hold `positions_lock` across `process()`→`append_mutation` — `append_mutation` acquires the `StateSnapshotLocks` bundle itself, and the launcher comment at `praxis/launcher.py:2664` explicitly notes wrapping it would re-acquire the non-reentrant `positions_lock` and self-deadlock.

To be precise, though: (a) removes the torn-read and narrows the window, but does **not** fully eliminate it. The capital mutation (`order_fill`) is under `capital_lock`, the dedup add under `positions_lock` — different domains — so a `checkpoint()` can still serialize capital-mutated-without-the-dedup-id in the `order_fill`→add gap and truncate the WAL. Fully closing it needs the dedup id recorded before the mutation under a lock held across both (record-before-mutate / `IN_PROGRESS/APPLIED`), which is the documented TD-086 remainder; the same residual already applies to capital-vs-`account_dust` and capital-vs-position (TD-048/075). This PR is still a strict improvement: pre-PR the dedup was process-local, so *every* restart dropped it and replayed-double-applied any persisted outcome; post-PR the id is durable via `append_mutation` in the common path, leaving only that narrow checkpoint-interleave-then-crash window.

Sweep: the two flagged dedup adds are the full population in this PR; `account_dust` writes were already correctly under `_positions_cm`, and the risk writes use `risk.lock` (pre-existing pattern, out of scope). No other instances.

nexus/infrastructure/praxis_connector/outcome_processor.py
blahmonkey: Addressed together with the L156 thread in 4cac5a5: this `processed_dust_close_ids` check+add is now folded into the existing `_positions_cm` block in `close_as_dust` (with the nested `_dedup_lock` keeping check+add atomic), so the dust dedup marker and its paired `account_dust` mutation serialize together under `positions_lock`. See the full analysis on the L156 thread (the residual checkpoint-interleave window is the TD-086 record-before-mutate remainder).

zero-bang (approved):

PR #101, opened 2026-06-22
feat: thread an injectable clock through CapitalController (v0.67.0)

blahmonkey:

## What does this PR change?

Adds an injectable `clock` to `CapitalController` so its time reads come from a supplied UTC source rather than the wall clock. This is the Nexus half of the Praxis replay engine (Vaquum/Praxis#122): a deterministic replay drives every component on one simulated cursor, so capital reservations must expire on simulated time, not wall time.

### Change

- `CapitalController.__init__` takes a keyword-only `clock: Callable[[], datetime]` defaulting to wall time, so the live path is unchanged.
- The three time reads — reservation creation (`check_and_reserve`), the expiry purge (`_purge_expired`), and `send_order` — now read `self._clock()`.

Without this, reservations created and checked within one real second would never appear expired even after simulated time advances past their TTL, so a replay's capital decisions would diverge from live.

## Tests

Two new tests (`TestInjectedClock` in `test_capital_controller.py`): a reservation's `expires_at` is computed from the injected clock, and `_purge_expired` honours it. `ruff` + `mypy` clean, the capital-controller suite passes. Bumps 0.66.0 → 0.67.0.

## Checklist


zero-bang (approved):
LGTM — clean, well-scoped change.

**Verified:**
- The three time reads the diff routes through `self._clock()` (`check_and_reserve`, `_purge_expired`, `send_order`) are the *only* `datetime.now` reads in `capital_controller.py` — coverage within the class is complete, nothing missed.
- All three reads happen under `self._lock`, so a replay cursor advanced between bars stays deterministic with no concurrency hazard.
- Default `clock=_utc_now` is a function reference as a default arg, so the live path is byte-for-byte unchanged. No existing caller needs updating (additive keyword-only param).
- Tests are meaningful and cover all three reads; the mutable-`moment` closure correctly models a cursor advancing past TTL.
- Consistent with the repo's existing injectable-clock convention (`now_fn` in `nexus/core/validator/intake_stage.py`).
- CI: all 5 checks green (CodeQL, Mypy, Ruff, Tests). The Copilot bot 'reviews' are quota-limit no-ops, nothing to address.

**Non-blocking observations (no action required):**
- Naming: this uses `clock`, while `intake_stage.py` uses `now_fn` for the same concept. Both are clear; just noting the two coexisting names in case you want to converge later.
- Forward-looking: deterministic replay isn't fully closed by this PR alone — other simulated-time-sensitive reads still hit the wall clock (e.g. `outcome_loop.py:372` staleness, `snapshot.py:52` retention cutoff, `shutdown_sequencer`/`state_store` timestamps). That's expected for 'phase one', but the next predictable failure mode is a replay whose capital decisions are deterministic while, say, outcome-staleness or snapshot-retention still keys off wall time — worth tracking which components remain wall-clock-bound until the cursor is threaded through all of them.

PR #104, opened 2026-07-12
feat: make ModeController the sole state.mode writer; persist health mode (v0.70.0)

blahmonkey:
### Change

Makes `ModeController` the sole writer of `state.mode` on the live path and persists the last health-derived mode across restart (Praxis WP-0008 follow-ups H1-route + H2).

- **Persist health mode (H2)**: new `InstanceState.health_mode`. `ModeController` reads/writes it instead of an in-memory `_health_mode`, and `reconcile` recommits from it directly (no more trigger-gated seeding). Fixes a health-derived `REDUCE_ONLY` being lost across a restart whenever a hold masked it as HALTED — clearing the hold then resumed ACTIVE instead of REDUCE_ONLY.
- **Sole writer (H1-route)**: `StartupSequencer._determine_mode` writes the boot decision into `state.health_mode`; the launcher's `reconcile` commits it. This also fixes a boot clobber where a no-health-data `REDUCE_ONLY` was overwritten with ACTIVE by `reconcile`. `ShutdownSequencer._halt_state_mode` places a dedicated `shutdown_hold` (ranked above manual/risk, un-liftable by a health tick) through an injected `ModeController`.
- `reconcile` clears any recovered `shutdown_hold` at boot — it is transient to the shutdown that placed it and the final checkpoint persists it, so a restart would otherwise boot HALTED-`shutdown` that operator-resume can't lift (caught in Greybeard review).

`wal_codec` persists `health_mode` and `shutdown_hold` with backward-compatible decode defaults.

## Tests

Full suite: 1515 passed. New coverage: shutdown-hold priority + un-liftable-by-health, persisted-health-mode-survives-restart, reconcile-commits-boot-mode-without-clobber, reconcile-clears-persisted-shutdown-hold (+ keeps a real manual hold), WAL round-trip + pre-field backward-compat, ShutdownSequencer routing (+ fallback). Two existing tests updated to the new mechanism.

## Follow-up

A one-line Praxis launcher change (pass `mode_controller=runtime.mode_controller` into the `ShutdownSequencer` construction) activates the shutdown-hold routing; it is backward-compatible (the sequencer falls back to a direct write when no controller is wired), so nothing breaks before it lands.

## Checklist


nexus/startup/shutdown_sequencer.py
```diff
@@ -204,13 +207,23 @@ def _halt_state_mode(self) -> None:
         any in-flight outcome dispatch's downstream order is rejected
         with `INTAKE_MODE_BLOCKS_ENTER` instead of leaking past
         `_dispatch_shutdown` to the venue.
+
+        Routes through the ModeController's shutdown hold when wired, so
+        the HALTED cannot be lifted by a concurrent health tick. Falls
+        back to a direct `state.mode` write only when no controller is
+        supplied (single-threaded test paths).
```
zero-bang: Minor doc-accuracy: this calls the fallback path "single-threaded test paths," but until the Praxis one-liner wires `mode_controller` in, production also constructs the sequencer with `mode_controller=None` and hits this direct-write branch — which is multi-threaded (HealthLoop running) and still exposes the resume-mid-shutdown race this PR is fixing. Consider rewording to "...when no controller is supplied (test paths and any launcher not yet wiring the controller)" so a future reader doesn't assume production is already covered.

zero-bang (dismissed):
Approve. This is a careful, well-scoped change and the sole-writer refactor holds up.

**What I verified**
- `reconcile` now recommits directly from persisted `state.health_mode` + recovered holds, and clears the transient `shutdown_hold` before recommitting — matching the Greybeard-review rationale (final checkpoint persists it, so a naive recover would boot HALTED-`shutdown` that operator-resume can't lift). Tests cover clear-vs-keep-manual.
- `_mode_source` / `any_active` shutdown priority is correct; a trigger-only recommit (already-HALTED manual → shutdown) updates the trigger without firing a spurious `on_halt`, and `test_shutdown_hold_takes_priority_over_manual_and_risk` locks that in.
- WAL decode defaults are sound and backward-compatible (missing `health_mode` → ACTIVE, missing `shutdown_hold` → inactive), and a *present-but-malformed* hold still fails decode.
- No consumer outside `mode_controller`/`wal_codec` enumerates the holds fields or switches on `mode.trigger` (only logging/change-detection treat it as opaque), so the new `'shutdown'` trigger value and the extra hold field introduce no exhaustiveness gaps. No import cycle from the new `enums` import in `instance_state`.

**Engineering notes (non-blocking)**
- On the real boot path `StartupSequencer._determine_mode` overwrites `state.health_mode` with a fresh boot evaluation before `reconcile` commits it. That means the *persisted* `health_mode` is effectively a transient default that the boot decision replaces every restart — the H2 persistence is really belt-and-suspenders behind the H1 field it introduces, and a persisted health-HALTED with no boot health data downgrades to REDUCE_ONLY on restart (strictly safer than the pre-PR ACTIVE, and self-corrects on the first HealthLoop tick). Worth a sentence in the `reconcile`/`_determine_mode` docstrings so a future reader doesn't expect the persisted value to win.
- The sole-writer correctness rests on the launcher ordering `reconcile` strictly *after* `start` (i.e. after `_determine_mode`). That ordering lives in Praxis, outside this diff — if it ever inverts, the boot mode written into `health_mode` would never reach `state.mode` until the first health tick. Worth an assertion/comment on the Praxis side when the follow-up lands.

**Forward-looking question**: once the Praxis one-liner wires `mode_controller` into the `ShutdownSequencer`, every graceful shutdown will drive `set_shutdown_halt` → `_recommit` → `on_halt('shutdown')`. If the launcher's `on_halt` pages/alerts operators, normal shutdowns will now emit a HALT alert. Is the shutdown source expected to be suppressed in the alerting callback?

nexus/core/domain/instance_state.py
```diff
@@ -65,6 +66,7 @@ class InstanceState:
     risk: RiskState = field(default_factory=RiskState)
     positions: dict[str, Position] = field(default_factory=dict)
     mode: ModeState = field(default_factory=ModeState)
+    health_mode: OperationalMode = OperationalMode.ACTIVE
     mode_holds: OperationalHolds = field(default_factory=OperationalHolds)
```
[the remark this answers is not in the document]
blahmonkey: Fixed in 59fe361. Documented `health_mode` in the `InstanceState` docstring Args: the last health-derived mode (the input `ModeController` arbitrates against the holds), persisted so a health-derived REDUCE_ONLY/HALTED survives a restart even while a hold masks it as HALTED — clearing the hold then restores it instead of resuming ACTIVE. Also clarified `mode` as the value `ModeController` commits from `health_mode` + `mode_holds`.

nexus/startup/shutdown_sequencer.py
blahmonkey: Good catch — fixed in 59fe361. Reworded the fallback note to drop the "single-threaded test paths" framing: it now reads "test paths, and any launcher not yet wiring the controller, where a concurrent HealthLoop tick can still lift the HALTED and re-expose the resume-mid-shutdown race this guards against." So a reader will not assume production is already covered until the Praxis one-liner (`mode_controller=runtime.mode_controller` into the `ShutdownSequencer` construction) lands.

zero-bang (approved):

